Issue #8, opened 2018-01-23
Billing Built-In


ohmybrew:
I'm going to build in the ability to charge for apps. I've done this many times, and figure its good to bake in as an (optional) thing for this package.

I plan on making it a separate middleware which anyone can attach to their routes. All settings for the billing including app name, pricing, trial period, etc, will be configurable.

Essentially the flow will be:

1. Shop gets authenticated with current middleware
2. Billing middleware will check if shop has paid (activation charge ID), or is grandfathered in (db flag)
3. If not paid:
   i. Redirect to Shopify billing screen
   ii. Shop accepts/declines and is redirected back to a billing controller
   iii. If accepted, charge is activated and stored in database for that shop, then redirected to home route.
   iiii. If declined, show error
4. If paid:
   i. Continue to route as normal

I'll be working on this in a new branch, with tests, and merge in once I feel its OK. Timeline for this is a bit sparse as I'm a tad busy these days so hang tight :)

ohmybrew:
Branch reference: [billable-feature](https://github.com/ohmybrew/laravel-shopify/tree/billable-feature).

ohmybrew:
So far:

+ Basics are setup
+ New configuration values in place
+ Shop model is modified to support billing
+ Middleware is implemented
+ Basic tests in place for middleware and new shop changes

Next:

+ Billing controller
  i. Handling redirect/display of Shopify billing page
  ii. Handle accept of billing
  iii. Handle decline of billing

ohmybrew:
Billing controller is now complete. Need to add tests for the processor action in the controller which completes things.

After that, some real-world tests.

ohmybrew:
Tests are now complete, code is also completed.

Working on updating some docs, and will begin real-world tests hopefully tomorrow, if not, early next week.

ohmybrew:
And we have success!

Install + Grandfathered mode: Working
Install + Pay mode: Working
Install + Decline mode: Working (shows error)

![screen-shot-2018-01-24-at-4 25 28-pm](https://user-images.githubusercontent.com/2420282/35354657-49c5a81a-0125-11e8-95bb-edd378b40e88.png)

I'll be auditing the code, pushing a release, and updating the wiki with information on enabling this feature over the next few days.

ohmybrew:
Based on tests (both code and real-world), we're good to merge in master.

From here, i'll update docs/wiki, do more tests, and release

ohmybrew:
Closing due to PR #10

PR #10, opened 2018-01-25
Billable feature

ohmybrew:
This PR will pull in the ability for baked-in billables. Single and recurring application charges for your apps easily with just setting up some configuration values.

See #8 issue for full details on the progress that brought this to life.

Will be updating docs here-after and releasing next week (possibly).


ohmybrew:
Waiting on CI checks to be extra sure it works across the multiple versions.

PR #17, opened 2018-03-11
Legacy mode

ohmybrew:
The ability to disable ESDK and use legacy mode.


ohmybrew:
Code tests pass, as well as real-world test.

PR #75, opened 2018-08-16
Fix bug with non-secure redirect URL.

paulcanning:
Fixes part of https://github.com/ohmybrew/laravel-shopify/issues/68

### DIFF
--- a/src/ShopifyApp/Traits/AuthControllerTrait.php
+++ b/src/ShopifyApp/Traits/AuthControllerTrait.php
@@ -59,7 +59,7 @@ trait AuthControllerTrait
         // Grab the authentication URL
         $authUrl = $api->getAuthUrl(
             config('shopify-app.api_scopes'),
-            url(config('shopify-app.api_redirect'))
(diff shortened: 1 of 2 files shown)


paulcanning:
I think the failed Travis CI build is due to a broken test for the BillingController, which I've not touched. Must be an existing issue.

ohmybrew:
@paulcanning A rebuild passes it, its a isolation issue on phpunit, I need to isolate one of the billing tests as it affects the database for the others (only for PHP 7.1 it seems).

Thanks for this. Its the other method I mentioned before. I do know for a fact, some people do build their test apps without SSL (just plain-ol HTTP), would it be worth it to do something like `url(config('shopify-app.api_redirect'), [], request()->secure())` ? This way it will generate https/http depending on the bool result of `secure()`... or should I simply drop support for that as app are "required" to be done through https regardless.

paulcanning:
Like you said, apps required HTTPS so I think it's safe enough to stick purely with secure_url()

PR #106, opened 2018-09-23
Allow for multiple after auth jobs

amosmos:


amosmos:
@tobiasdalhof - I agree - that's a great idea. Will try to make it.

amosmos:
@tobiasdalhof - actually - even if the parameter is only 1 job and not an array, it is still a php array.

Any idea how can this be checked?

tobiasdalhof:
Checking `config('shopify-app.after_authenticate_job')[0]` should work

tobiasdalhof:
Something like this comes to my mind

```php
$afterAuthJobConfig = config('shopify-app.after_authenticate_job');
$runSingleJob = !empty($afterAuthJobConfig['job']);

if ($runSingleJob) {
    // Run job
    return
}

foreach ($afterAuthJobConfig as $job) {
    // Run job
}
```

amosmos:
The failed travis test is is now not necessary as the jobs are now installed with a foreach loop, which will not run if there are now jobs to run.

ohmybrew:
Thanks for this. Will review after my flights but from the phone looks all good :) the failed test, wifi here is bad, but it's probably the billing controller. There's a test in there which I haven't figured out how to solve, it fails only sometimes due to an order of operations thing, usually restart of the Travis test makes it pass. Either way I'll check it out!

amosmos:
Thanks! Happy to help with this great project!

This is the failed test:
`OhMyBrew\ShopifyApp\Test\Controllers\AuthControllerTest::testAfterAuthenticateDoesNotFireForNoConfig
Failed asserting that true is false.`

And it's ok that it is failing because indeed now the afterauth method does not return false if the parameter is empty, and that is because it runs in a foreach loop which will simply not run if the array is empty.

Thanks again!
Amos

ohmybrew:
Looks fine to me. Will merge in and kill that test and should be fine since its a loop anyways.

PR #108, opened 2018-09-25
Re-organize Suggestion

amosmos:


src/ShopifyApp/Traits/AuthControllerTrait.php
```diff
@@ -207,4 +185,28 @@ protected function returnTo()
...
     }
+
+    /**
+     * Fires ש job.
+     *
+     * @param array $config The job's configuration
+     *
+     * @return bool
+     */
+    protected function fireJob($job)
```
ohmybrew: I believe this will fail since $job needs to be $config. Did you update the tests as well?

ohmybrew:
@amosmos Have you seen my comment above?

ohmybrew:
Will close, you're free to reopen if you get to it!

src/ShopifyApp/Traits/AuthControllerTrait.php
amosmos: Not sure I understand your comment, I simply changed some structure, didn't change any logic. I moved the function to be a class function instead of an anonymous function inside the afterAuthenticateJob function.

amosmos:
Hi, see my answer to your comment above, not sure how to reopen the request :-)

amosmos:
Hi @ohmybrew - did you have a chance to look at my answer?

Thanks!

ohmybrew:
On mobile so can't see fully but if I remember, what you moved, you didn't carry a variable or two over which would cause it to fail. However this logic is now all moved to a service so you would need to update the PR to reflect the last release as the code in the PR no longer exists in the spot its in :)

amosmos:
Ah I see, ok will look for that :-) THANKS!

PR #129, opened 2018-10-17
Get shop model from config

### LINKED ISSUE #128: Resolve Shop model from config in AppUninstalledJob
Currently, in AppUninstalledJob.php, base model of package (Shop) is directly being used to find the shop.

`use OhMyBrew\ShopifyApp\Models\Shop;`

Kindly resolve it using the `shopify-app.php` inside config as we can override it in most of the cases.

Problem faced: Events on the extended Shop model are not fired, for example Shop Delete event.

tuimz:
This closes #128


ohmybrew:
Thank you!

PR #140, opened 2018-11-15
Adjustment to job classes for loose checking in constructor

awebartisan:
While using ShopModelTrait, base Shop model or its child class is not passed in constructor of two jobs, `WebhookInstaller` and `ScriptTagInstaller`, so we can remove the type hinting from constructor to solve this issue.


ohmybrew:
thanks!

PR #155, opened 2018-12-20
load migrations from the package

amosmos:
instead of exporting them to the app's migration folder

### DIFF
--- a/src/ShopifyApp/ShopifyAppProvider.php
+++ b/src/ShopifyApp/ShopifyAppProvider.php
@@ -35,9 +35,7 @@ class ShopifyAppProvider extends ServiceProvider
         ], 'config');
 
         // Database migrations
-        $this->publishes([
-            __DIR__.'/resources/database/migrations' => "{$this->app->databasePath()}/migrations",
-        ], 'migrations');
+        $this->loadMigrationsFrom(__DIR__.'/resources/database/migrations');
 
         // Job publish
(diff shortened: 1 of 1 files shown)


ohmybrew:
I think we need to export them based on my reading from packages for Laravel. This is to allow the package user to modify the migrations before committing them. Thoughts?

ohmybrew:
Interesting, I'm open to it, ill pull it in!

ncpope:
I would suggest an optional flag that would allow the user to either copy the migrations to the migrations directory or to use auto migration loading with `loadMigrationsFrom`.

There could be an .env setting that could control it. Something like `SHOPIFY_MANUAL_MIGRATIONS=false` or similar could be the default.

If the developer wants to use the default migrations they wouldn't have to do anything by default. 

However if they want to change the migrations they can use the current flow and copy the tables to their migrations directory and modify them as necessary.

This could allow for a simple if/else check when running the migrations: 

```php
/**
 * Perform post-registration booting of services.
 *
 * @return void
 */
public function boot()
{
    if (!config('shopify-app.manual_migrations')) {
        $this->loadMigrationsFrom(__DIR__.'/path/to/migrations');
    }
}
```

ncpope:
The reason I suggest this is so that it doesn't break existing modifications to the migrations and it allow the package to be flexible for a wider variety of use cases moving forwards.

PR #160, opened 2018-12-28
Automatic billing plan selection based on a shops Shopify plan

ncpope:
Enables the automatic on-install charge based on a shops Shopify plan.

To test first publish the migrations and run the new migration

```bash
 php artisan vendor:publish --tag=migrations && php artisan migrate
```

Then add the following line to the .env
```.env
SHOPIFY_BILLING_TIERED_PRICING_ENABLED=true
```

Add the following to config/shopify-app.php
```php

    /*
    |--------------------------------------------------------------------------
    | Tiered Billing
    |--------------------------------------------------------------------------
    |
    | Enable billing tiers based on a shops Shopify plan.
    |
    */

    'billing_tiered_pricing_enabled' => (bool) env('SHOPIFY_BILLING_TIERED_PRICING_ENABLED', false),
```

And finally create a new plan
```sql
INSERT INTO `plans` (`type`, `name`, `price`, `capped_amount`, `terms`, `trial_days`, `test`, `on_install`, `created_at`, `updated_at`, `shop_plan`)
VALUES
	(1, 'Special Affiliate Plan (Test Mode)', 0.99, NULL, NULL, 7, 1, 0, '2018-12-28 12:58:53', NULL, 'affiliate');
```


ohmybrew:
Thats a neat idea. I think we can scope the plan to grab [here](https://github.com/ohmybrew/laravel-shopify/blob/master/src/ShopifyApp/Traits/BillingControllerTrait.php#L31). This wouldn't require a change in the middleware, and we just grab based on their shopify plan IF the setting is enabled, what do you think?

ncpope:
@ohmybrew  That's a good idea. I think that makes more sense than what I was doing. I'll submit a new pull request when I get a chance in the next day or two

ARKKYN:
@ncpope  any updates on this

ncpope:
@ARKKYN I forgot all about this pull request. I will adapt the suggestion from @ohmybrew then submit another pull request.

ncpope:
No time to support this request.

PR #168, opened 2019-01-20
make asset publishing unique to this app

amosmos:


ohmybrew:
Interesting. Is this something other packages are doing?

amosmos:
Yes, of course, even Laravel's own packages use namedspaced tags for their asset publishing.

The reason I noticed it was that so it happens that there's another builtin package in Laravel that uses the tag 'config' and every time I publish your package assets, that one is published too, and I had no way to stop this.

ohmybrew:
Ah well in that case I can adjust the tags.

On 1/28/19 12:02, Amos Shacham wrote:

Yes, of course, even Laravel's own packages use namedspaced tags for their asset publishing.

The reason I noticed it was that so it happens that there's another builtin package in Laravel that uses the tag 'config' and every time I publish your package assets, that one is published too, and I had no way to stop this.

PR #173, opened 2019-01-25
Fix/Check if is already a JsonResponse

ericnkatz:
An additional fix in regards to https://github.com/ohmybrew/laravel-shopify/issues/164#issuecomment-457462347

> `return response()->json(Model::all());`
> 
> Yields a response that is text/html; charset=UTF-8 but looks like
> 
> ```
> HTTP/1.1 200 OK
> Cache-Control: no-cache, private
> Content-Type:  application/json
> Date:          Fri, 25 Jan 2019 05:39:47 GMT
> 
> []
> ```


ericnkatz:
Sounds good, thanks for all your work on this project. It's been a delight to use. 👍

Issue #181, opened 2019-02-18
Migration Refresh issue


ilamp:
Executing `php artisan migrate:refresh` causes following issue:

`Illuminate\Database\QueryException  : SQLSTATE[HY000]: General error: 1828 Cannot drop column 'plan_id': needed in a foreign key constraint 'charges_plan_id_foreign' (SQL: alter table `charges` drop `plan_id`)`

ilamp:
I have the bug fix for this. Please allow me access so that I can submit a pull request. @ohmybrew

ohmybrew:
Hi! Please open a pull request with the code change and I can merge it in.

ilamp:
Doing it right now.

ilamp:
I see that the travis throwing an error:

`SQLite doesn't support dropping foreign keys (you would need to re-create the table).`

Do you want me to look into that?

ilamp:
@ohmybrew Hey Tyler, pull request should be passing the tests now. Please let me know if you wanted me to change anything else.
Thanks
Daljeet

PR #182, opened 2019-02-19
Migration Refresh issue #181

ilamp:


ohmybrew:
Thanks! Getting an error on my side that DB is unresolved. Can you add a use statement for it? It'll fix then :)

ilamp:
I believe I already added that in the next commit. Let me double check.

ohmybrew:
Great I'll merge in the afternoon :sunglasses:

ilamp:
Sounds good

PR #194, opened 2019-03-11
Fix incorrect DocBlock

bilfeldt:
Fixing incorrect variable name in DocBlock . Used the same naming as in `src/ShopifyApp/Jobs/AppUninstalledJob.php`


ohmybrew:
Thanks!

PR #240, opened 2019-05-16
Configurable myshopify domain

dhritimanrick:
myshopify domain can be managed through environment variables


ohmybrew:
Thank you, will merge.

PR #255, opened 2019-06-01
Add debug configuration and add logging

StefanNeuser:
As we already discussed here are the first lines of logging to get more information what is happening in die middleware. 

You can enable and disable logging within the shopify-app.php config file.


StefanNeuser:
@ohmybrew are you using phpStorm or any other IDE with a code-formatter setting that you can share?

ohmybrew:
Hi Stefan. 

I am using VSCode with phpcs.

ohmybrew:
Instead of doing the inline check in places. How about adding a Log function to ShopifyApp. Keep the code dry.

Then you can simply do: ShopifyApp::Log(message here) anywhere without always checking the config.

PR #261, opened 2019-06-06
Add remainingDaysAfterCancel for Charge Model

StefanNeuser:
Hi all,

to check how many days a left in a canceled recurring plan i added this method. 

Please check if that makes also sense for you.

@ohmybrew do you have a code-formatter configuration that you can share? PHPStorm? VsCode? Sublime?

### DIFF
--- a/src/ShopifyApp/Models/Charge.php
+++ b/src/ShopifyApp/Models/Charge.php
@@ -210,7 +210,7 @@ class Charge extends Model
         $remainingDays = 30 - Carbon::parse($this->activated_on)->diffInDays(Carbon::today()) % 30;
 
         // the cancellation-date is before a new period of 30 days begins
-        if ($cancelledDate->diffInDays(Carbon::today()) != 0 && $remainingDays == 30)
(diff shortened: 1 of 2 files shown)


src/ShopifyApp/Models/Charge.php
```diff
@@ -169,30 +171,65 @@ public function remainingTrialDays()
...
+            return false;
+        }
+
+
+        $cancelledDate = Carbon::parse($this->cancelled_on);
+
+        $periods = (int)(Carbon::parse($this->activated_on)->diffInDays(Carbon::today()) / 30);
+        $pastDaysInPeriod = Carbon::parse($this->activated_on)->addDays(30 * $periods)->diffInDays(Carbon::today());
+
+        if ($pastDaysInPeriod == 0 && $cancelledDate->lt(Carbon::today()))
```
StefanNeuser: @ohmybrew i dont know if this special-case makes sense right now. Maybe somebody can validate this. The idea is when the user cancel on the date when a new 30 days period begins. than a new period of 30 days should be active before cancel comes into play

PR #262, opened 2019-06-06
Helper in Charge Model for RecurringCharges (Periods)

StefanNeuser:
this methods should help to check if a when a RecurringCharge period begins and ends


ohmybrew:
@StefanNeuser Why did you close this? Not yet ready?

PR #263, opened 2019-06-09
Charges Model helper methods for recurring-charges

StefanNeuser:
Hi all,

i added a test and also a bunch if new helper method for recurring-charges. 

The PR looks little bit messy because mit IDE is reformatting it when i edit it. Is there a code-style that i cant use?

Best Regards from Germany,

Stefan


ohmybrew:
Thanks! Looks good to me, ill have style ci fix the formatting on merge.

PR #283, opened 2019-07-17
Prefixed routes

dellow:
Adds the config option to add a prefix on the URLs.

I wanted to use the home route of my apps site as a landing page for my app and have all the laravel-shopify stuff under `/app`. This pull request allows that.

I haven't added any unit tests because it's just a simple change. I have been running my app on it though and it all works.

The only caveat is the user must set the existing SHOPIFY_API_REDIRECT to match. So my config options would be:

    SHOPIFY_APP_PREFIX=app // New config option.
    SHOPIFY_API_REDIRECT=app/authenticate // Existing config option, but matches above.


ohmybrew:
Thank you for this. A wiki article for this would be helpful as well, I can write one.

ohmybrew:
[WIKI entry](https://github.com/ohmybrew/laravel-shopify/wiki/Prefixing-Routes)

ohmybrew:
Thanks for this, will be in new release.

dellow:
No problem. I also found the `SHOPIFY_BILLING_REDIRECT` needs to be updated with the same prefix.

Haven't found any issues at all since changing this.

dellow:
Have updated the Wiki regarding the `SHOPIFY_BILLING_REDIRECT` change.

PR #293, opened 2019-08-01
simplify auth and bugfix

StefanNeuser:
move ->setShop(...) outside the `elseif` with `$session->isValid()` check

### DIFF
--- a/src/ShopifyApp/Middleware/AuthShop.php
+++ b/src/ShopifyApp/Middleware/AuthShop.php
@@ -21,14 +21,15 @@ class AuthShop
      * Handle an incoming request.
      *
      * @param \Illuminate\Http\Request $request
-     * @param \Closure                 $next
+     * @param \Closure $next
      *
      * @return mixed
      */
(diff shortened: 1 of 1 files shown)


ohmybrew:
Thanks! Slightly modded it to add in an HMAC check for session updates. Realized this is needed.

Test: Shop is brand new, does not exist
Expected: Redirect to install process. Full redirect.
Result: Success
 
Test: No existing session
Expected: Session created and HMAC validated.
Result: Success

Test: With existing shop in session, try to load app in a different shop
Expected: Since shops do not match, session must be updated. No full redirect.
Result: Success

Test: With existing shop in session, try and load app (admin link) in a different shop
Expected: Since shops do not match, session must be updated. No full redirect.
Result: Success


----


Doing a couple updated unit tests... will get back to you shortly!

ohmybrew:
Edit: Units came out fine, so did physical tests, would you like to pull from master @StefanNeuser and see if it solves the issues you've been facing?

PR #297, opened 2019-08-09
try to get shop-domain from header

StefanNeuser:
because we have this issue https://github.com/ohmybrew/laravel-shopify/issues/295 i try also to get the shopify_domain from the header. in our app we set the header by 

`window.axios.defaults.headers.common['shop'] = document.head.querySelector('meta[name="shop"]').content`

### DIFF
--- a/src/ShopifyApp/Middleware/AuthShop.php
+++ b/src/ShopifyApp/Middleware/AuthShop.php
@@ -48,7 +48,7 @@ class AuthShop
         $session = new ShopSession();
 
         // Grab the shop's myshopify domain from query or session
-        $shopDomainParam = $request->get('shop');
+        $shopDomainParam = $request->get('shop') ?? $request->header('shop');
         $shopDomainSession = $session->getDomain();
         $shopDomain = ShopifyApp::sanitizeShopDomain($shopDomainParam ?? $shopDomainSession);
 


darrynten:
This is very dangerous and would allow an attacker to load any shop by spoofing the header

StefanNeuser:
@darrynten , no it isnt because of the $session->isValid() check in the AuthShop.php middleware

src/ShopifyApp/Middleware/AuthShop.php
```diff
@@ -48,7 +48,7 @@ protected function validateShop(Request $request)
         $session = new ShopSession();
 
         // Grab the shop's myshopify domain from query or session
-        $shopDomainParam = $request->get('shop');
+        $shopDomainParam = $request->get('shop') ?? $request->header('shop');
```
darrynten: 1. This would allow injection and account takeover via request header spoofing
2. `Shop:` is not a good header name, it would have to be something like X-Shopify-Shop or something (might be reserved)

StefanNeuser: 1. no because of the `$session->isValid()` in the same file

2. agree we can rename it

darrynten: 3. This would mean all devs would have to make this change. This is not the right way to implement such a change because it means it's broken by default

StefanNeuser: https://stackoverflow.com/questions/187655/are-https-headers-encrypted

darrynten: 4. Blocks users from being able to simply POST a `form` element - this would mean everyone would have to use a library to post forms instead of just using forms how they should be

darrynten: > And by the way : https://stackoverflow.com/questions/187655/are-https-headers-encrypted

That doesn't have anything to do with anything though, it's not about the value being exposed it's about being able to arbitrarily send anything in that value

StefanNeuser: 4. in laravel $request->get() is also fetching post params, you can set the shop are are hidden field like you do with csrf-token from laravel

StefanNeuser: > 3. This would mean all devs would have to make this change. This is not the right way to implement such a change because it means it's broken by default

thats correct in generell but the only think we do here is to also allow to pass the shop in the header instead only check for cookie/get/post param at all

darrynten: @StefanNeuser see https://github.com/ohmybrew/laravel-shopify/pull/298 for my fix

StefanNeuser:
@darrynten your fix is the better one: https://github.com/ohmybrew/laravel-shopify/pull/298

PR #298, opened 2019-08-09
Fix for session clashes

darrynten:
Sessions were clashing when a user was managing the same app for multiple stores in the same browser.

Both stores have the same session due to them both using the app domain in the same browser.

This was leading to store A overwriting store B or vice-versa.

A solution for normal req/resp apps (i.e. not SPAs) is to check the referer header to see exactly which store we're dealing with.

Because the referer header is spoofable we make sure that there is a valid HMAC signature present in the referer.

If there is a shop domain with a valid HMAC signature in the referer header then we overwrite the shopDomain variable to make sure the correct store is loaded.

Resolves https://github.com/ohmybrew/laravel-shopify/issues/295

### DIFF
--- a/src/ShopifyApp/Middleware/AuthShop.php
+++ b/src/ShopifyApp/Middleware/AuthShop.php
@@ -110,7 +110,7 @@ class AuthShop
      *
      * @return bool|string
      */
-    protected function getShopDomain(Request $request, ShopSession $session)
+    private function getShopDomain(Request $request, ShopSession $session)
     {
         // Query variable is highest priority
         $shopDomainParam = $request->get('shop');
@@ -146,7 +146,7 @@ class AuthShop
      *
      * @return bool|string
      */
-    protected function getRefererDomain(Request $request)
+    private function getRefererDomain(Request $request)
     {
         // Extract the referer
         $referer = $request->header('referer');
--- a/src/ShopifyApp/Middleware/AuthShop.php
+++ b/src/ShopifyApp/Middleware/AuthShop.php
@@ -39,6 +39,10 @@ class AuthShop
     /**
      * Get the referer shopify domain from the request and validate.
      *
+     * It is dangerous to blindly trust user input so we need to
+     * check and confirm the validity upfront before we return the
+     * value to anything.
+     *
      * @param \Illuminate\Http\Request $request
      *
      * @return bool|string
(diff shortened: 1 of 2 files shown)


StefanNeuser:
@darrynten, your fix is not working for us. we tried it multiple times now. the referrer seems to be empty after the second click inside the app. we are using https://github.com/inertiajs/inertia-react

darrynten:
@StefanNeuser I think we're trying to solve 2 different problems 😆

src/ShopifyApp/Middleware/AuthShop.php
```diff
@@ -52,6 +52,12 @@ protected function validateShop(Request $request)
         $shopDomainSession = $session->getDomain();
         $shopDomain = ShopifyApp::sanitizeShopDomain($shopDomainParam ?? $shopDomainSession);
 
+        // See issue https://github.com/ohmybrew/laravel-shopify/issues/295
+        parse_str(parse_url($request->header('referer'), PHP_URL_QUERY), $refererQueryParams);
+        if (isset($refererQueryParams['shop']) && $shopDomain !== $refererQueryParams['shop'] && ShopifyApp::api()->verifyRequest($refererQueryParams)) {
```
darrynten: @ohmybrew if I don't verify the params then I can modify the referer domain to be the other shops domain and it loads the modified domain (basically make the problem exhibit itself again) and makes it possible for an attacker to gain control of a shop through IDOR.

I will adjust the code to make it only verify if it really and truly needs to and add the tests when they're done.

OWASP A4 - [IDOR - Insecure Direct Object References](https://www.owasp.org/index.php/Top_10_2013-A4-Insecure_Direct_Object_References)

There would have been concerns about performance if verifying the request through the API called the server, but I checked the code in `OhMyBrew/BasicShopifyAPI.php` and it just does a signature check - it doesn't make any network requests.

It's important to not trust data received from the client, and to sanitise and check actions like this.

darrynten:
> @darrynten, your fix is not working for us. we tried it multiple times now. the referrer seems to be empty after the second click inside the app. we are using https://github.com/inertiajs/inertia-react
> 
> But your change makes also sense. What do you think about to add
> 
> `$shopDomainParam = $request->get('shop') ?? $request->header('shop');`

This PR is for normal req/resp type pages that use a form, not SPAs

ohmybrew:
The refactor is great. Much more testable. Is it good to merge in or you still need time with it?

darrynten:
I'm sorted, if you're happy with it then it's good to go.

mansoorbilalrana:
@ohmybrew @darrynten   Still facing the shop session issue between two stores on the same browser.
Store A shows data from Store B. Please suggest. 
image 
image

Kyon147:
Are you on v10.1 as I'm sure this was looked at again.

mansoorbilalrana:
@Kyon147  I am using "version": "v10.1.0".

Kyon147:
Are your routes behind the auth.shop middleware?

mansoorbilalrana:
Yes, all routes are behind the auth.shop middleware.

Kyon147:
It might be worth debugging the authentication route on your local, so you can see why the session is not being refreshed on the second store.

I remember needing to do the same before v9 as I had constant issues with the session and it pulling through the wrong data.

PR #299, opened 2019-08-09
Fix for session clashes also for Single-Page-Apps

StefanNeuser:
According to https://github.com/ohmybrew/laravel-shopify/pull/298 and the PR that i closed before for the issue https://github.com/ohmybrew/laravel-shopify/issues/295

### DIFF
--- a/src/ShopifyApp/Middleware/AuthShop.php
+++ b/src/ShopifyApp/Middleware/AuthShop.php
@@ -48,9 +48,16 @@ class AuthShop
         $session = new ShopSession();
 
         // Grab the shop's myshopify domain from query or session
-        $shopDomainParam = $request->get('shop') ?? $request->header('shop');
+        // For SPA's we need X-Shop-Domain : See issue https://github.com/ohmybrew/laravel-shopify/issues/295
+        $shopDomainParam = $request->get('shop') ?? $request->header('X-Shop-Domain');
         $shopDomainSession = $session->getDomain();
         $shopDomain = ShopifyApp::sanitizeShopDomain($shopDomainParam ?? $shopDomainSession);
+
+        // See issue https://github.com/ohmybrew/laravel-shopify/issues/295
+        parse_str(parse_url($request->header('referer'), PHP_URL_QUERY), $refererQueryParams);
+        if (isset($refererQueryParams['shop']) && $shopDomain !== $refererQueryParams['shop'] && ShopifyApp::api()->verifyRequest($refererQueryParams)) {
+            $shopDomain = $refererQueryParams['shop'];
+        }
 
         // Get the shop based on domain and update the session service
         $shopModel = Config::get('shopify-app.shop_model');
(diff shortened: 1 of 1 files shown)


ohmybrew:
@StefanNeuser @darrynten Can't tell on mobile which one of you did what :D so just tagging here to discuss.

Looks good at a glance, can you explain why this is needed... also the verifyRequest, should not be needed on every request? If the session is bad, going to a full or partial authentication flow will already verify the request.. unless I'm missing something (https://github.com/ohmybrew/laravel-shopify/blob/master/src/ShopifyApp/Requests/AuthShop.php#L46).

src/ShopifyApp/Middleware/AuthShop.php
```diff
@@ -48,10 +48,17 @@ protected function validateShop(Request $request)
         $session = new ShopSession();
 
         // Grab the shop's myshopify domain from query or session
-        $shopDomainParam = $request->get('shop');
+        // For SPA's we need X-Shop-Domain : See issue https://github.com/ohmybrew/laravel-shopify/issues/295
+        $shopDomainParam = $request->get('shop') ?? $request->header('X-Shop-Domain');
```
darrynten: If you really want to send the domain in the headers you should really also send the HMAC and verify if to prevent spoofing.

So send the both the shop and the signature retrieved from the headers to the `api()->verify request()` call with something like `['shop' => 'example.com', 'hmac' => 'abc123']`

darrynten:
@ohmybrew take a look at my explanation on #298 for the reason behind my change.

src/ShopifyApp/Middleware/AuthShop.php
```diff
@@ -48,10 +48,17 @@ protected function validateShop(Request $request)
...
         $shopDomainSession = $session->getDomain();
         $shopDomain = ShopifyApp::sanitizeShopDomain($shopDomainParam ?? $shopDomainSession);
 
+        // See issue https://github.com/ohmybrew/laravel-shopify/issues/295
+        parse_str(parse_url($request->header('referer'), PHP_URL_QUERY), $refererQueryParams);
+        if (isset($refererQueryParams['shop']) && $shopDomain !== $refererQueryParams['shop'] && ShopifyApp::api()->verifyRequest($refererQueryParams)) {
+            $shopDomain = $refererQueryParams['shop'];
+        }
+
```
darrynten: You shouldn't combine PRs like this.

1 feature/fix per PR should be followed.

This now contains my semi-unrelated referer fix as well as a change to allow modifying of store to be loaded via request headers.

darrynten: Please remove this functionality from this PR as it exists in another PR. Leave fixes where they belongs - in their own PRs.

If changes are made in either PR then both would be out of sync.

For example, I am busy adding tests to make sure coverage in my PR goes up.

We should also be having the conversation about re-verification on the PR where this code is.

Thanks! 😎

darrynten: I have completely refactored my PR and this is now completely different.

Please remove all of these old lines of mine from your PR.

Thank you!

darrynten:
@StefanNeuser I've opened some other PRs too with some refactoring - what I will do is after those are in I will be able to make an implementation that will allow an SPA to announce itself more securely.

I need to do this anyway, and this would achieve your objective in #299.

StefanNeuser:
@darrynten , i agree and removed the code from your PR (patch)

PR #300, opened 2019-08-10
Add PHP 7.3 to travis

darrynten:
Run the tests against PHP 7.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: php
 php:
   - 7.1
   - 7.2
+  - 7.3
 
 env:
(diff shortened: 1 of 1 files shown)


darrynten:
@ohmybrew something is up with PHP 7.3 but not too sure what it is - have added it to travis to highlight the issue in coveralls.

ohmybrew:
Did you find the issue? I can dig too

darrynten:
Would be great. I'm going to dig a little deeper too as soon as I have time because we're running on 7.3

The beauty of tests 😎

PR #305, opened 2019-08-12
Security Fix for A4 IDOR Vulnerability

darrynten:
### Problem

It was possible for an attacker to successfully load a different stores information into memory via insecure direct object references.

You could simply change the domain in the get variables to another valid store and the application would successfully load their shop.

To fix this we add verification at the point of entry, and before we load the target shops sensitive model data into memory.

### Explanation

- [Line 52](https://github.com/ohmybrew/laravel-shopify/blob/master/src/ShopifyApp/Middleware/AuthShop.php#L52) loads the shop specified by the user
- [Line 59](https://github.com/ohmybrew/laravel-shopify/blob/master/src/ShopifyApp/Middleware/AuthShop.php#L59) then loads this value, without any validation, into the session

From that point on the arbitrary target shop is loaded. It's important that user input gets sanitised and checked before doing any processing on it.

### Changes

- Require signature verification before loading any shops
- Refactor loading from query variables
- Refactor loading from headers
- Allow submission of details via POST as well as GET
- Use correct header variable when loading from headers
- Add load shop from query param signature validation
- Add load shop from header param signature validation
- Add load shop from referer signature failed exception
- Updated all tests

### Impact

This requires no changes for shops that:

1. Use GET variables like normal (still passes as long as no fiddling)
2. Rely on the Referer variable, like a normal req/resp site (throws exception if fiddling)
3. Fall back to Session value

This only requires changes for shops that:

1. Want to use the new loading domain from headers functionality - they must include the signature details in their request (unreleased feature)
2. Use a GET variable without a signature (a hack that should not be allowed to happen)

#### For SPAs

Track and append 4x values on requests:

- `X-Shop-Domain` - the domain of the target store
- `X-Shop-Signature` - the signature (a.k.a. hmac)
- `X-Shop-Time` - the timestamp used in the creation of the signature
- `X-Shop-Code` - the code used in the creation of the signature

If these values have been tampered with an exception will throw.

#### For GET/POST

Exactly like before, in the same way that Shopify prescribes.

- `shop` - the domain of the target store
- `hmac` - the signature
- `timestamp` - the timestamp
- `code` - the code

If these values have been tampered with an exception will throw.

#### For Referer

- If the referer does not have all 4 values in the GET vars the loading via referer will skip.
- If the referer has all 4 values but the verification fails an exception will throw.

#### For Session

If none of the above is present then the shop domain will be loaded from session.

This is an acceptable risk because to take over a store an attacker would have to steal another stores cookie (unlike before where they could just modify a query param).

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,3 +57,4 @@ This project is released under the MIT [license](https://github.com/ohmybrew/lar
 ## Misc
 
 I develop this package in my spare time, with a busy family/work life like many of you! So, I would like to thank everyone who's helped me out from submitting PRs, to assisting on issues, and plain using the package (I hope its useful). Cheers.
+
--- a/src/ShopifyApp/Middleware/AuthShop.php
+++ b/src/ShopifyApp/Middleware/AuthShop.php
@@ -154,23 +154,22 @@ class AuthShop
             return false;
         }
 
-        $signatureHeaderParam = $request->input('hmac');
-        $timeHeaderParam = $request->input('timestamp');
-        $codeHeaderParam = $request->input('code');
+        $signature = $request->input('hmac');
+        $timestamp = $request->input('timestamp');
+        $code = $request->input('code');
 
         // Make sure there is no param spoofing attempt
         if (ShopifyApp::api()->verifyRequest([
             'shop' => $shop,
-            'hmac' => $signatureHeaderParam,
-            'timestamp' => $timeHeaderParam,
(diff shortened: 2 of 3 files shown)


darrynten:
@ohmybrew a more secure implementation with verification

@StefanNeuser passing domain of store in a secure way for your requirements

ohmybrew:
Great breakdown, this'll make a good wiki entry to... pulled in.

PR #309, opened 2019-08-13
Fix Signature Optional Params

### LINKED ISSUE #306: AuthShop is failing
Hi @darrynten , i have a issue with the last change. Auth is not working anymore because the exception is thrown everytime and it seems there is no `code` in the input when i open the app from the shopify admin. Do you have any idea?

https://github.com/ohmybrew/laravel-shopify/blob/5b68ee73272dbde83316b6cc09e3eaea4059af0d/src/ShopifyApp/Middleware/AuthShop.php#L165

i guess there is a missing check like here:
https://github.com/ohmybrew/laravel-shopify/blob/5b68ee73272dbde83316b6cc09e3eaea4059af0d/src/ShopifyApp/Middleware/AuthShop.php#L203

Here you can see what shopify is sending in the first request:

Screenshot 2019-08-12 at 19 59 45

@ohmybrew i guess that the `code` within the request from shopify-admin is only set for a `full-auth` flow. So thats why the `verifyRequest` is failing for us. Maybe i did something wrong?

darrynten:
## Problem

The signature verification was assuming the `code` param would always be present. This was making non-full flow requests that do not have this present fail.

Additionally, there are optional params that could form part of the signature, like `locale` and `state`.

## Solution

Take the optional params into account when verifying the signature.

Resolves #306 

### Header Requests

There are 3 required and 3 additional optional headers depending on how the signature has been generated.

No impact as feature is unreleased, but documentation should be updated to reflect these new requirements.

#### Required

- X-Shop-Domain
- X-Shop-Signature
- X-Shop-Timestamp

#### Optional

- X-Shop-Locale
- X-Shop-Code
- X-Shop-State

### Referer Detection

Takes additional optionals into account. No impact as feature is unreleased.

### GET/POST Requests

No impact, but will fix issues when code is not present, or when locale or state have been used in generation of the signature.

## Changes

- Update domain retrieval methods
- Update signature verification param optionality
- Add lots of tests to cover as many combinations as possible


darrynten:
@StefanNeuser this should fix the problems you were having. Please test and LMK if all is working as expected for you.

Thanks!

StefanNeuser:
@darrynten thank you very much this solves the problem for everybody not just me ;)

darrynten:
@StefanNeuser it's my pleasure, happy to help.

Hopefully this will provide everyone with a safe enough way to load the correct store when dealing with the same app in multiple shops in the same browser.

I think I covered all of the possible params that can be part of the signature, but in case I missed something you now know where the cause could be.

Cheers!

Unrelated: @ohmybrew has coveralls stopped?

StefanNeuser:
> I think I covered all of the possible params that can be part of the signature, but in case I missed something you now know where the cause could be.

@darrynten Did you find a documentation section within the shopify-docs where all the hmac relevant params are documented?

darrynten:
@StefanNeuser I found some info [here](https://help.shopify.com/en/api/getting-started/authentication/oauth)

They do say not to rely on those values as they're subject to change, but it seems to be working fine for now.

ohmybrew:
Both are merged in, feel free to give it a go and see if it solves the root issue :) ill test it out shortly as well.

ohmybrew:
I haven't found an issue with my physical tests... I think we're good for a release, will confirm tonight.

StefanNeuser:
@ohmybrew in your video it seems you are using the `?shop=...` param. Remove it or otherwise use the hmac params. There is also another issue open where i replied with the same solution as here. Hope that fit the needs

PR #310, opened 2019-08-14
Add parameter for hmac verifcation

StefanNeuser:
When you come from a shopify-admin link shopify is sending a `id` param. That was missing in the newly added hmac verification.


StefanNeuser:
fyi: @darrynten

darrynten:
👍 I approve! Nice catch :)

PR #318, opened 2019-08-23
Update app installation page

ncpope:
Updated app login view to more closely match the official shopify gem login view: https://github.com/Shopify/shopify_app/blob/master/app/views/shopify_app/sessions/new.html.erb

### CURRENT SCREEN

Screen Shot 2019-08-23 at 13 29 39

### UPDATED SCREEN

Screen Shot 2019-08-23 at 13 30 08


StefanNeuser:
nice shot!

PR #335, opened 2019-09-30
Bug fix: `appbridge_version` incorrectly casting to boolean

jedimdan:
Spotted a typo in the new `appbridge_version` config incorrectly casting to a boolean, resulting in the URL to app bridge always returning a 1 or a 0 rather than the actual config value.


ohmybrew:
Thanks

Issue #327, opened 2019-09-03
AuthShop middleware needs to handle exceptions gracefully


ncpope:
*For bug reporting only! If you're posting a feature request or discussion, please ignore.*

## Expected Behavior

If the shopify_domain is not set on a route protected by the auth.shop middleware there should either be a redirect to the /login route or a specific exception thrown.

## Current Behavior

Currently there is just a general Exception being thrown.

### Steps to Reproduce

See my repo where I followed installation steps: https://github.com/ncpope/tester/commits/master

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Package Version: ^9.2
* Laravel Version: 5.8.*
* PHP Version: PHP 7.3.4
* Using a toolset (Vagrant):

### Failure Logs

<img width="1664" alt="Screen Shot 2019-09-03 at 10 58 50" src="https://user-images.githubusercontent.com/1903724/64154589-dc8a0000-ce39-11e9-97ad-4f7ac3af9ba6.png">

ncpope:
Nevermind, I was using an outdated version of the package. Closing....

jedimdan:
Hi @ncpope, which version did you get to that resolved this problem? I'm on 9.2 and it's still throwing this exception for me. From what I can see, `AuthShop` will still throw an exception when there's no shopify domain, and this exception isn't caught anywhere.

ncpope:
@jedimdan you're right, I just took a quick glance yesterday as I ran into the bug on a new projects using 9.2.0 and I assumed that the changes would somehow handle redirecting to the `/login` route. The current behavior of throwing an exception is not ideal as the middleware will just fail with an unhandled exception being thrown.

https://github.com/ohmybrew/laravel-shopify/blob/554a121983e7f3af8c79d5e3e34a0b928f09b34c/src/ShopifyApp/Middleware/AuthShop.php#L136

As far as I can tell I don't see a reason for this line to throw an exception. I think it should just redirect to the `/login` route.

I am going to re-open this issue to encourage discussion. Any clarification behind this would be appreciated.

StefanNeuser:
Hi @ncpope i really understand your issue i thought about the same. What you can do for now is listening in `App\Exceptions\Handler` for  the `MissingShopDomainException` or other Exceptions. 

You can check it with `if( $exception instanceof ...)` in the render method and redirect or return with some output. 

Hope this helps

jedimdan:
[...]

PR #347, opened 2019-10-15
Typo: changed `shopify-app.debug` key to `app.debug`

jedimdan:
I believe @ohmybrew meant to write `app.debug` since `shopify-app.debug` doesn't actually exist. (Reference: https://github.com/ohmybrew/laravel-shopify/issues/327#issuecomment-536872852)

### DIFF
--- a/src/ShopifyApp/Exceptions/BaseException.php
+++ b/src/ShopifyApp/Exceptions/BaseException.php
@@ -22,7 +22,7 @@ abstract class BaseException extends Exception
      */
     public function render(Request $request)
     {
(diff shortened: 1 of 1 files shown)


jedimdan:
@ohmybrew thanks for taking in my feedback from #327! I was actually waiting to hear your thoughts about it so that I can assist by doing a PR, since you're the maintainer and we want to honor your opinion first. I think some of us are open to help carry out the change for you so that it doesn't all have to be on your shoulders.

ohmybrew:
Ah thanks, missed this one. I'll have to update the test too. Thanks!

PR #350, opened 2019-10-30
UsageCharge Service - return Charge Model instead of boolean

StefanNeuser:
Hi @ohmybrew , we extended the charges table for usageCharges and saw that the save method just return the result of the eloquent save operation. To get the charge model after saving we changed this. Please take a look and maybe discuss :)

Have a nive day! Stefan


ohmybrew:
Sorry, I've been super busy :(

I'm perfectly fine with this - makes sense! Will pull in for next minor release

Issue #277, opened 2019-07-09
I am facing an issue on embedded app in iphone safari in login section


azarrar:
When i connect to myApp using shopify admin section, App is working but when i use login section of laravel it is showing 419 error of laravel only in apple safari. but working find in all browser chorme, firefox.

azarrar:
resolved by enabling cookies in safari.

bilfeldt:
The problem is third party cookies which is actually disabled in Safari per default. Check your browser here: [https://www.whatismybrowser.com/detect/are-third-party-cookies-enabled](https://www.whatismybrowser.com/detect/are-third-party-cookies-enabled)

Actually, this is can be a problem for many users since Safari is pretty widely used and this is a default setting.

@ohmybrew and @azarrar I suggest re-opening this issue. 

The only way to solve this would probably be to rely on sessions instead of third party cookies. I am not sure it that can be done @ohmybrew? But at least we can add a middleware that can detect this so that it's possible to show a more detailed page with information about how to enable this in Safari.

What do you think?

ohmybrew:
To my knowledge there is nothing I can do on my end @bilfeldt :( While the idea of showing a message is nice, I think its something not needed, I don't want to have a bunch of code for every situation like this.

bilfeldt:
@ohmybrew This is caused by the Oauth process. There is a question about this on the Shopify forum [here](https://community.shopify.com/c/Shopify-APIs-SDKs/How-to-enable-third-party-cookies-embedded-apps/td-p/439425).

Basically this 419 error is caused by the app setting the cookie from within the iFrame which is blocked by browsers like Safari as part of ITP 2.0, since the cookie domain does not correspond to the parent domain - see Github discussion [here](https://github.com/Shopify/shopify_app/issues/533#issuecomment-361683084)

Shopify has created an entire page with recommended flow for handling ITP 2.0, which can be seen [here](https://help.shopify.com/en/api/guides/itp-impact). This is the recommended flow:

![image](https://user-images.githubusercontent.com/19682/45030452-da6c4600-b019-11e8-8dcf-68b0aadd17a2.png)

I think it would be a great addition to the plugin, to handle this for Safari (preferably a setting which browsers to do this for) since Safari is a widely used browser and it seems this might be implemented by other browsers as well (I noted that it seems that some users are experiencing this on some versions of Firefox as well, but havn't confirmed if it is the same error).

[...]

PR #375, opened 2020-01-20
Middleware that sets cookies to samesite=none

### LINKED ISSUE #367: Cookies on embedded apps blocked by Chrome 80
## Expected Behavior

Successfully authenticate embedded app on Chome 80.

## Current Behavior

Shopify store goes into a redirect loop as the cookies from the app are rejected by Chrome 80.

### Steps to Reproduce

1. [Install Chrome Beta](https://www.google.com/chrome/beta/)
2. Launch the app within the Shopify store
3. The app would not authenticate successfully.

### Context

Come February 2020, Chrome 80 will be released and will begin rejecting cookies from a third party site that is not set to `SameSite=None; Secure`. (Ref: [Chromium blog](https://blog.chromium.org/2019/10/developers-get-ready-for-new.html)) According to Google, Mozilla and Microsoft will be making the same changes in future versions of their browser.

As embedded apps on Shopify are considered third-party, if the app's cookie is not set to `SameSite=None; Secure`, the behaviour described above would occur. To further complexify things, Safari on macOS 10.14 and iOS 12 (and [some others](https://www.chromium.org/updates/same-site/incompatible-clients)) would behave incorrectly when a cookie with `SameSite=None; Secure` is submitted. As such, the app needs to not send that flag to some of these browsers. In other words, simply setting the cookies config in Laravel would not be sufficient.

While this is a Laravel thing more than an issue with this package (Laravel supports setting these flags on cookies via `session.php`), all projects that use this package to develop embedded apps on Shopify would encounter this issue. So perhaps it could be something this package could handle?

Thought to bring this up as a topic of discussion. I'm open to contribute code to this but would love to hear your thoughts first, especially from @ohmybrew. 

**Additional Reference:** [Shopify article about this issue](https://help.shopify.com/en/api/guides/samesite-cookies)

jedimdan:
Attempt to fix #367.

What I'm not sure is the strategy to insert this middleware. Should this be included in the install docs for developers to include into the global middleware list? The consideration is that this only applies for requests that are within an embedded app.

Also, the other part I'm not sure is if you are okay with a new dependency, seeing that this library has been very clean from other dependencies so far.

**Based on Shopify's implementation:** https://github.com/Shopify/shopify_app/pull/851

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "php": ">=7.1.0",
         "laravel/framework": "~5.6 | ^6.0",
         "ohmybrew/basic-shopify-api": "^6.0",
+        "jenssegers/agent": "^2.6",
         "doctrine/dbal": "~2.5"
     },
     "require-dev": {
--- a/src/ShopifyApp/Middleware/SameSiteNone.php
+++ b/src/ShopifyApp/Middleware/SameSiteNone.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace OhMyBrew\ShopifyApp\Middleware;
+
+use Closure;
+use Jenssegers\Agent\Agent;
+
+class SameSiteNone
+{
+    /**
+     * Sets SameSite=None while checking for incompatible browsers
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $userAgent = $request->userAgent();
+        $headers = $request->header();
+
+        // set SameSite none to supported browsers only
+        if ($this->isBrowserSameSiteNoneCompatible($userAgent, $headers)) {
+            config(['session.secure' => true]);
+            config(['session.same_site' => 'none']);
+        } // else continue with original session config
+
+        return $next($request);
+    }
+
+    /**
+     * Checks using the user agent if the browser is compatible with SameSite None.
+     * If a null is passed for any of the variables,
+     * @param  string|null  $userAgent
+     * @param  array|null  $headers
+     * @return boolean True if compatible false if not
+     */
+    public function isBrowserSameSiteNoneCompatible($userAgent, $headers)
+    {
+        // manually initialise the parser so that this function is testable
+        $agent = new Agent;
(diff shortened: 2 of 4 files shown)


jedimdan:
@darrynten I've got this working middleware that I already use in my production. I just need help figuring out what's the most appropriate way to make it available in this package. Do you have any suggestions?

darrynten:
I've been looking at this and there is no need to add any code to fix this issue.

The change you mentioned in #277 works perfectly.

For others who may need this info, the change needs to be done in the host laravel app that is using this.

You make the following changes in `config/session.php`

Change:

```
    'secure' => env('SESSION_SECURE_COOKIE', false),
```

to

```
    'secure' => env('SESSION_SECURE_COOKIE', true),
```

Change:

```
    'same_site' => null,
```

to

```
    'same_site' => 'none',
```

After that's done then the expected policies are in place and Shopify is happy with the update.

This means that the changes proposed in this PR are redundant, and I would vote to close the PR without merging.

jedimdan:
Hi @darrynten the additional changes are necessary because of the browser compatibility issues. It's not sufficient to simply change the cookie settings. 

I've explained this in detail in #367. The challenge is that the browsers that have issues are quite recent (iOS 12 and macOS 10.14) so it's quite critical that the compatibility issues are covered. This code is based on Shopify's changes to their own Rails library as well.

jedimdan:
Regarding compatibility issues please read [this article](https://www.chromium.org/updates/same-site/incompatible-clients). This cookie completely breaks the app on Safari for iOS 12 and macOS 10.14 so it needs to not change in those browsers.

darrynten:
I'm actively working on this, I will update as soon as possible

jedimdan:
> I'm actively working on this, I will update as soon as possible

What's your game plan?

jedimdan:
@darrynten I also recommend editing your recommendation above on changing the sessions config as doing so will cause the app to unexpectedly break on Safari for macOS 10.14, which I believe is still heavily used.

darrynten:
Please see #382 it resolves the problem cleanly on the Service level instead of on the Middleware level

jedimdan:
Closing this PR in favour of #382

PR #382, opened 2020-01-31
Chrome Cookies

darrynten:
From ~~4th of Feb~~ 17th of Feb Shopify Apps must all send SameSite=none cookies with the secure flag set.

This is because Chrome 80+ has this as a hard requirement in order for embedded apps to continue to work.

This PR is based on the initial work performed by @jedimdan in PR #375 

The key difference here is that the check has been moved to the ShopSession Service level, instead of being on the Middleware level, and the tests are performed on actual requests, not just the check method.

Additionally, all derived Shopify Host addresses will be https instead of http (this is a hard requirement for secure cookies to work).

This patch makes the need to edit the `config/sessions.php` file in the host application redundant. Changing those values can be harmful as it would set those flags for incompatible browsers and result in broken apps for those users.

Resolves #367 
~~#277~~ (See comment from @jedimdan for why this is not actually resolved)

## References

[Shopify Announcement](https://help.shopify.com/en/api/guides/samesite-cookies)
[Known Incompatible Clients](https://www.chromium.org/updates/same-site/incompatible-clients)

PR #386, opened 2020-02-04
Bug fix for incorrect SameSite=none behaviour on some platforms.

jedimdan:
A bug was discovered from PR #382 in `ShopSession.php`'s new `getBrowserDetails()` and `getPlatformDetails()` method that causes the app to not return SameSite=none on some platforms. This is likely caused by the fact that some platform's version numbers do not have a minor number. 

The proposed fix is to rely on `Jenssegers\Agent`'s built-in functions that can generate a float version number from any user-agent.

Reference: https://github.com/ohmybrew/laravel-shopify/pull/382#issuecomment-581620833_

Other fixes include:
- A mistake in the unit-test, where a compatible UCBrowser user-agent wound up in the incompatible list (confusion probably caused because both "Chrome" and "UCBrowser" is mentioned in the user-agent).
- Flipping the logic from a whitelist to a blacklist, since `SameSite=None` is expected to be a norm on more browsers moving forward ([Refer to this comment](https://github.com/ohmybrew/laravel-shopify/pull/386#issuecomment-581756542))
- Updated Travis Laravel version constraint to properly support Laravel's new Semantic Versioning

### DIFF
--- a/src/ShopifyApp/Services/ShopSession.php
+++ b/src/ShopifyApp/Services/ShopSession.php
@@ -299,19 +299,19 @@ class ShopSession
                 $compatible = true;
             }
 
-            if ($this->agent->is('iOS') && $platform['float'] > 12) {
+            if ($this->agent->is('iOS') && $platform['float'] >= 13) {
                 $compatible = true;
             }
 
             if ($this->agent->is('OS X') &&
                 ($this->agent->is('Safari') && !$this->agent->is('iOS')) &&
-                $platform['float'] > 10.14
+                $platform['float'] >= 10.15
             ) {
                 $compatible = true;
             }
 
             if ($this->agent->is('UCBrowser') &&
-                $browser['float'] > 12.13
+                $browser['float'] >= 12.132
             ) {
                 $compatible = true;
             }
--- a/src/ShopifyApp/Services/ShopSession.php
+++ b/src/ShopifyApp/Services/ShopSession.php
@@ -295,11 +295,11 @@ class ShopSession
             $browser = $this->getBrowserDetails();
             $platform = $this->getPlatformDetails();
 
-            if ($this->agent->is('Chrome') && $browser['major'] >= 67) {
+            if ($this->agent->is('Chrome') && $browser['float'] >= 67) {
                 $compatible = true;
             }
 
-            if ($this->agent->is('iOS') && $platform['major'] > 12) {
+            if ($this->agent->is('iOS') && $platform['float'] > 12) {
                 $compatible = true;
             }
 
@@ -329,13 +329,10 @@ class ShopSession
      */
     private function getBrowserDetails()
     {
-        $version = $this->agent->version($this->agent->browser());
-        $pieces = explode('.', str_replace('_', '.', $version));
+        $version = $this->agent->version($this->agent->browser(), Agent::VERSION_TYPE_FLOAT);
 
         return [
-            'major' => $pieces[0],
-            'minor' => $pieces[1],
-            'float' => (float) sprintf('%s.%s', $pieces[0], $pieces[1]),
+            'float' => $version,
         ];
     }
 
@@ -346,13 +343,10 @@ class ShopSession
      */
     private function getPlatformDetails()
     {
-        $version = $this->agent->version($this->agent->platform());
-        $pieces = explode('.', str_replace('_', '.', $version));
+        $version = $this->agent->version($this->agent->platform(), Agent::VERSION_TYPE_FLOAT);
 
         return [
-            'major' => $pieces[0],
-            'minor' => $pieces[1],
-            'float' => (float) sprintf('%s.%s', $pieces[0], $pieces[1]),
+            'float' => $version,
         ];
     }
 }
--- a/src/ShopifyApp/Services/ShopSession.php
+++ b/src/ShopifyApp/Services/ShopSession.php
@@ -295,7 +295,7 @@ class ShopSession
             $browser = $this->getBrowserDetails();
             $platform = $this->getPlatformDetails();
 
-            if ($this->agent->is('Chrome') && $browser['float'] < 67) {
+            if ($this->agent->browser() == 'Chrome' && $browser['float'] < 67) {
                 $compatible = false;
             }
 
@@ -304,13 +304,13 @@ class ShopSession
             }
 
             if ($this->agent->is('OS X') &&
-                ($this->agent->is('Safari') && !$this->agent->is('iOS')) &&
+                ($this->agent->browser() == 'Safari' && !$this->agent->is('iOS')) &&
                 $platform['float'] < 10.15
             ) {
                 $compatible = false;
             }
 
-            if ($this->agent->is('UCBrowser') &&
+            if ($this->agent->browser() == 'UCBrowser' &&
                 $browser['float'] < 12.132
             ) {
                 $compatible = false;
--- a/src/ShopifyApp/Services/ShopSession.php
+++ b/src/ShopifyApp/Services/ShopSession.php
@@ -291,35 +291,31 @@ class ShopSession
 
         $this->agent = new Agent();
 
-        try {
-            $browser = $this->getBrowserDetails();
-            $platform = $this->getPlatformDetails();
+        $browser = $this->getBrowserDetails();
+        $platform = $this->getPlatformDetails();
 
-            if ($this->agent->browser() == 'Chrome' && $browser['float'] < 67) {
-                $compatible = false;
-            }
-
-            if ($this->agent->is('iOS') && $platform['float'] < 13) {
-                $compatible = false;
-            }
-
-            if ($this->agent->is('OS X') &&
-                ($this->agent->browser() == 'Safari' && !$this->agent->is('iOS')) &&
-                $platform['float'] < 10.15
-            ) {
-                $compatible = false;
-            }
-
-            if ($this->agent->browser() == 'UCBrowser' &&
-                $browser['float'] < 12.132
-            ) {
-                $compatible = false;
-            }
-
-            return $compatible;
-        } catch (\Exception $e) {
-            return true;
+        if ($this->agent->browser() == 'Chrome' && $browser['float'] < 67) {
+            $compatible = false;
         }
+
+        if ($this->agent->is('iOS') && $platform['float'] < 13) {
+            $compatible = false;
+        }
+
+        if ($this->agent->is('OS X') &&
+            ($this->agent->browser() == 'Safari' && !$this->agent->is('iOS')) &&
+            $platform['float'] < 10.15
+        ) {
+            $compatible = false;
+        }
+
+        if ($this->agent->browser() == 'UCBrowser' &&
+            $browser['float'] < 12.132
+        ) {
+            $compatible = false;
+        }
+
+        return $compatible;
     }
 
     /**
--- a/src/ShopifyApp/Services/ShopSession.php
+++ b/src/ShopifyApp/Services/ShopSession.php
@@ -287,7 +287,7 @@ class ShopSession
      */
     private function checkSameSiteNoneCompatible()
     {
-        $compatible = false;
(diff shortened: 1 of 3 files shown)


jedimdan:
@mferrario @muhammadasfar Could you guys share some user agents that you've seen fail on your app so we can add it to the tests to confirm that this fixes the issues you're seeing?

jedimdan:
@darrynten Would like your feedback on this PR

jedimdan:
@darrynten also right now the logic is a whitelist of compatible browsers rather than a blacklist of incompatible browsers. Based on the Chrome team's writeup, they suggest the blacklist as all other browsers should not have an issue with the new cookie flag. We also do expect more browsers to follow suit in the future, so having a blacklist might make more sense since there won't be a need to keep updating the code every time a new browser vendor adopts the same requirement.

muhammadasfar:
@jedimdan  

I am using chrome 80 Beta and user agents is

`Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.78 Safari/537.36`

**Domain:** https://myprojectname.test

In my case this function is return `Undefined offset: 1` as show explanation in screenshot
samesite-none-issue-src

And now I am trying this version release on my live app domain for confirmation.

mferrario:
> @mferrario @muhammadasfar Could you guys share some user agents that you've seen fail on your app so we can add it to the tests to confirm that this fixes the issues you're seeing?

@jedimdan here are two examples:

My Code:
`
 $agent = new Agent();            
 $browser = $agent->browser();
 $version = $agent->version($browser);            
 $platform = $agent->platform();
 $platform_version = $agent->version($platform);      
 $robot = $agent->robot();

// print
print "Visitor: ".$_SERVER['SERVER_NAME']." | $browser - $version | $platform - $platform_version | Robot: $robot";
`
// Actual values
Visitor: 52.0.4.244 | Mozilla - | - | Robot: Zgrab
secure-stage.jerseywatch.net | Mozilla - | - | Robot: Nimbostratus-Bot

jedimdan:
@mferrario Could you try printing `$_SERVER['HTTP_USER_AGENT']` instead? As the error we are trying to fix has to do with deciphering the actual user agent string.

src/ShopifyApp/Services/ShopSession.php
```diff
@@ -287,38 +287,38 @@ public function setCookiePolicy()
      */
     private function checkSameSiteNoneCompatible()
     {
-        $compatible = false;
+        $compatible = true;
```
darrynten: Why would you switch this?

If you start with the default of compatible then random browsers that do not pass the 'fail' checks will be marked as compatible and will break.

For example, browsing from lynx in the terminal, using the playstation or switch browser, random internet connected fridges and cars etc.

It's better to start off as incompatible, and then only mark the ones that are definitely compatible.

src/ShopifyApp/Services/ShopSession.php
```diff
@@ -329,13 +329,10 @@ private function checkSameSiteNoneCompatible()
      */
     private function getBrowserDetails()
     {
-        $version = $this->agent->version($this->agent->browser());
-        $pieces = explode('.', str_replace('_', '.', $version));
+        $version = $this->agent->version($this->agent->browser(), Agent::VERSION_TYPE_FLOAT);
```
darrynten: This is great, I had no idea getting a float was possible through the library

tests/Services/ShopSessionTest.php
```diff
@@ -122,21 +122,23 @@ public function testCanForget()
     public function testSameSiteCookie()
     {
         $incompatibleUserAgents = [
+            'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.78 Safari/537.36',
             'Mozilla/5.0 (iPhone; CPU iPhone OS 12_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) GSA/87.0.279142407 Mobile/15E148 Safari/605.1',
             'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.1.2 Safari/605.1.15',
             'Mozilla/5.0 (Linux; U; Android 7.0; en-US; SM-G935F Build/NRD90M) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 UCBrowser/11.3.8.976 U3/0.8.0 Mobile Safari/534.30',
             'UCWEB/2.0 (Java; U; MIDP-2.0; en-US; generic) U2/1.0.0 UCBrowser/9.5.0.449 U2/1.0.0 Mobile',
-            'Mozilla/5.0 (Linux; U; Android 9; zh-CN; Nokia X7 Build/PPR1.180610.011) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/57.0.2987.108 UCBrowser/12.14.0.1020 Mobile Safari/537.36',
```
darrynten: Any particular reason this one was removed?

tests/Services/ShopSessionTest.php
```diff
@@ -151,14 +153,26 @@ public function testSameSiteCookie()
...
+            $this->assertEquals('none', $response->baseResponse->headers->getCookies()[0]->getSameSite());
+            $this->assertTrue($response->baseResponse->headers->getCookies()[0]->isSecure());
         }
 
         foreach ($incompatibleUserAgents as $agent) {
+            // reset sessions before each test
+            config([
+                'session.secure'    => false,
+                'session.same_site' => null,
+            ]);
```
darrynten: Nice catch

src/ShopifyApp/Services/ShopSession.php
jedimdan: So far the examples from both Chrome (https://www.chromium.org/updates/same-site/incompatible-clients) and Shopify (https://github.com/shopify/shopify_app/pull/851/) default as compatible and only mark as incompatible in the scenario that it is known to not work. I feel that it's wiser to adhere to their strategy than to be counter to it.

I believe the reasoning is because older browsers will ignore this flag and treat it as the default setting for a cookie, and managing a list of "compatible" browsers is only going to be harder to maintain with time as more and more browsers (such as Firefox and MS Edge) begin to enforce stricter cookie protections. In fact, the list of incompatible browsers Google has identified are browsers that behave unexpectedly when the flag is passed. That is why iOS 12 is incompatible, but iOS 11 is compatible (it simply ignores the flag).

tests/Services/ShopSessionTest.php
jedimdan: This wasn't removed. It got moved to the compatible list. This is a user agent for UBrowser 12.14, not Chrome 57.

jedimdan: By the way, this is also the reason why I've changed the browser detection from `$agent->is()` to `$agent->browser() ==`. This particular user agent returns true for `$agent->is('chrome')` while `$agent->browser()` returns `UCBrowser`, which is why the browser version ends up (confusingly during my debugging) returning `12.14` instead of `57.0`

jedimdan:
@darrynten by the way, apologies for not spotting and highlighting some of these in your original PR.

ohmybrew:
Feel free to tag me once you feel its out of draft, and I'll gladly push a patch release.

mferrario:
> @mferrario Could you try printing `$_SERVER['HTTP_USER_AGENT']` instead? As the error we are trying to fix has to do with deciphering the actual user agent string.

Here are some new examples:

//PRINT
print "Visitor: ".$_SERVER['SERVER_NAME']." :: ".$_SERVER['HTTP_USER_AGENT']." | $browser - $version | $platform - $platform_version | Robot: $robot";

// OUTPUT
Visitor: 52.0.4.244 :: Mozilla/5.0 (X11; Linux x86_64; rv:52.0) Gecko/20100101 Firefox/52.0 | Firefox - 52.0 | Linux - | Robot:
Visitor: 54.152.130.178 :: Mozilla/5.0 zgrab/0.x | Mozilla - | - | Robot: Zgrab
Visitor: www.doingaworldofgood.org :: Mozilla/5.0 (compatible; Baiduspider/2.0; +http://www.baidu.com/search/spider.html) | Mozilla - | - | Robot: Baiduspider
Visitor: 54.152.130.178 :: Mozilla/5.0 NetSeen/1.0 | Mozilla - | - | Robot:
Visitor: stage.jerseywatch.com :: Mozilla/5.0 (compatible; Nimbostratus-Bot/v1.3.2; http://cloudsystemnetworks.com) | Mozilla - | - | Robot: Nimbostratus-Bot
Visitor: example.com :: AWS Security Scanner | - | - | Robot: AWS Security Scanner
Visitor: [::ffff:a9fe:a9fe] :: AWS Security Scanner | - | - | Robot: AWS Security Scanner
Visitor: 169.254.169.254 :: AWS Security Scanner | - | - | Robot: AWS Security Scanner
Visitor: 52.0.4.244 :: checks | - | - | Robot:
Visitor: 54.152.130.178 :: Mozilla/5.0 NetSeen/1.0 | Mozilla - | - | Robot:
Visitor: 169.254.169.254 :: AWS Security Scanner | - | - | Robot: AWS Security Scanner
Visitor: [::ffff:a9fe:a9fe] :: AWS Security Scanner | - | - | Robot: AWS Security Scanner

jedimdan:
@darrynten One more thing I wanted to ask: with the change to relying on the Agent class' built-in float versions, I assume the try-catch is no longer necessary? Also, let me know if my answers to your comments are acceptable.

src/ShopifyApp/Services/ShopSession.php
jedimdan: Actually, now that I think about it, technically the only browser that supports `SameSite=None` is Chrome. Neither UCBrowser nor Safari support that flag even now. So by that theory, only Chrome should be marked as compatible. 

So for Google to recommend only blacklisting browsers that are known to break a site with that flag, it means that all other browsers should be assumed will ignore that flag and considered "compatible".

darrynten: I wouldn't take that example from Chrome as gospel. They will of course assume their way is correct and forget about things like crawlers, scripts that use cookie jars etc. SameSite=none isn't even a standard yet, it's still in draft.

Perhaps they're trying to use their large market share to force the change to become standard. If everyone starts sending SameSite=none; Secure back to all browsers it's easier for them to get the standard to pass.

I personally believe it's best to assume that everything is _not_ compatible with SameSite=none, and add browsers that are compatible with this new Chrome standard over time (if any).

However, I do get what you're saying about obscure browsers possibly just ignoring the SameSite=none flag and treating it the same as SameSite=lax

That being said, I'm happy with either approach, as long as it ensures all of our apps continue to work.

tests/Services/ShopSessionTest.php
darrynten: Interesting.

Wow, this change is a real headache, and I think a large part of the internet is going to 'break' when Chrome 80 rolls out.

darrynten:
@jedimdan yes, it seems the try/catch will not be needed anymore

src/ShopifyApp/Services/ShopSession.php
jedimdan: I get where you're coming from @darrynten. 

For the sake of our apps continuing to work without having to play catch-up as Edge and Firefox adopts stricter `SameSite=None` rules (both vendors have already agreed to follow suit, but have no schedule yet), I think this is the better option. Besides, since Shopify's official Rails library has adopted the black-list strategy (https://github.com/Shopify/shopify_app/pull/851), if we get this wrong, so would have all of our fellow Rails peers. So I think we are safe ;)

jedimdan:
Thanks, @ohmybrew, this PR is ready. Thanks @darrynten for the review!

ohmybrew:
Thanks!
   
   Will do this tonight.

PR #430, opened 2020-03-23
Only process needed request values

andrewscofield:
I was getting a PHP compile error on `implode` because I was sending other data through the an API route. Meaning I had payload data in my request and this function was not processing it correctly.

**Note:** I'm not entirely sure that `$shopify_values` actually contains all the values needed.

### DIFF
--- a/src/ShopifyApp/Http/Middleware/AuthShopify.php
+++ b/src/ShopifyApp/Http/Middleware/AuthShopify.php
@@ -236,9 +236,12 @@ class AuthShopify
             // GET/POST
             DataSource::INPUT()->toNative() => function () use ($request): array {
                 // Verify
+                $shopify_values = array('shop', 'hmac', 'timestamp', 'timestamp');
                 $verify = [];
                 foreach ($request->all() as $key => $value) {
-                    $verify[$key] = is_array($value) ? '["'.implode('", "', $value).'"]' : $value;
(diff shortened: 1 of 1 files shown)


osiset:
The problem with $request->all() is it returns both post and get doesn't it?

I think maybe if we change it to $request->query() would be better as we would only grab the query vars, not post?

andrewscofield:
Yea that is the issue I'm running into. Although, I don't know why anybody would use it, I think query vars can also be an array, so it would also through the same error if it ever encountered a complex set of query vars. 

https://www.php.net/manual/en/function.http-build-query.php  (Example 3)

osiset:
I'll switch it to use query only... But I'll dig through old issues to be sure.. pretty sure it came up for a reason just can't remember whyOn Mar. 25, 2020 00:00, andrewscofield <notifications@github.com> wrote:
Yea that is the issue I'm running into. Although, I don't know why anybody would use it, I think query vars can also be an array, so it would also through the same error if it ever encountered a complex set of query vars.
https://www.php.net/manual/en/function.http-build-query.php  (Example 3)

osiset:
@andrewscofield Sorry, I've been super tied up. Are you able to mod your PR to use query and not all?

andrewscofield:
Got it changed and its working great for me.

osiset:
Awesome, thanks man

PR #431, opened 2020-03-25
Fix counting remaining trial days when recharge cancelled billing

aepnat:
Charge model using SoftDelete. When shop reinstalling app and recharge billing with trial plan, we should find latest charge(include trashed) data for counting remaining of trial days.


osiset:
Thank you, merged.

Issue #419, opened 2020-03-18
Fresh install is broken on Laravel 7.x


ncpope:
## Expected Behavior

On a fresh install you should be able to install osiset/laravel-shopify without any errors using the install steps in the wiki.

## Current Behavior

There is a conflict during the migration process. 

## Failure Information

php artisan vendor:publish --tag=shopify-migrations && php artisan migrate
Unable to locate publishable resources.
Publishing complete.
Migration table created successfully.
Migrating: 2014_10_12_000000_create_users_table
Migrated:  2014_10_12_000000_create_users_table (0.02 seconds)
Migrating: 2019_08_19_000000_create_failed_jobs_table
Migrated:  2019_08_19_000000_create_failed_jobs_table (0.01 seconds)
Migrating: 2020_01_29_010501_create_plans_table
Migrated:  2020_01_29_010501_create_plans_table (0.01 seconds)
Migrating: 2020_01_29_230905_create_shops_table
Migrated:  2020_01_29_230905_create_shops_table (0.02 seconds)
Migrating: 2020_01_29_231006_create_charges_table

   Illuminate\Database\QueryException

  SQLSTATE[HY000]: General error: 1215 Cannot add foreign key constraint (SQL: alter table `charges` add constraint `charges_user_id_foreign` foreign key (`user_id`) references `users` (`id`) on delete cascade)

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:669
    665|         // If an exception occurs when attempting to run a query, we'll format the error
    666|         // message to include the bindings with SQL, which will make this exception a
    667|         // lot more helpful to the developer instead of just the database's errors.
    668|         catch (Exception $e) {
  > 669|             throw new QueryException(
    670|                 $query, $this->prepareBindings($bindings), $e
    671|             );
    672|         }
    673|

      +34 vendor frames
  35  artisan:37
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))

### Steps to Reproduce

Just install the package as normal on a fresh install using 7.x

### How to fix

Option 1. Set the users id to increment using integers and not bigint
- $table->increments('id');

Option 2. Set the user_id type to a big int on the charges migration
- Current : $table->integer('user_id')->unsigned();
- Suggestion : $table->bigInteger('user_id')->unsigned();

mtownsend5512:
[...]

PR #449, opened 2020-04-11
correct version check for Laravel 5.8 introduction of bigIncrements

aaronlp:
Laravel changed the user_id column to be a big increments in v5.8 and this bug was raised and fixed to cater for it:
https://github.com/osiset/laravel-shopify/issues/419

However, the version check excludes v5.8 when it should be included. This PR fixes the conditional version check.


osiset:
Thanks... Laravel has been tricky with this undocumented change.

PR #469, opened 2020-04-30
Parse query strings the same as Shopify to support array parameters

joelvh:
Shopify does not parse the query string of a request with array parameters the same way as Laravel. While Laravel and most web frameworks parse `name[]=value` to `['name' => ['value']]`, Shopify parses it to `['name[]' => 'value']`. (When there are multiple parameters with the same name, Shopify will put the values in an array.)

The reason for this is shown in the [Shopify docs](https://shopify.dev/tutorials/display-data-on-an-online-store-with-an-application-proxy-app-extension): they use [`Rack::Util.parse_query`](https://github.com/rack/rack/blob/f9ad97fd69a6b3616d0a99e6bedcfb9de2f81f6c/lib/rack/query_parser.rb#L36).

```ruby
require 'rack/utils'

# Use request.query_string in rails
query_string = "extra=1&extra=2&shop=shop-name.myshopify.com&path_prefix=%2Fapps%2Fawesome_reviews&timestamp=1317327555&signature=a9718877bea71c2484f91608a7eaea1532bdf71f5c56825065fa4ccabe549ef3"

query_hash = Rack::Utils.parse_query(query_string)
# => {
#   "extra" => ["1", "2"],
#   "shop" => "shop-name.myshopify.com",
#   "path_prefix" => "/apps/awesome_reviews",
#   "timestamp" => "1317327555",
#   "signature" => "a9718877bea71c2484f91608a7eaea1532bdf71f5c56825065fa4ccabe549ef3",
# }
```

We've been running into issues with array parameters, and decided to port the Ruby code to PHP to ensure the exact same query string parsing as Shopify uses to create the HMAC signature.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -1,5 +1,6 @@
 # [...document.getElementsByClassName('contrib-person')].map(n => n.getElementsByClassName('avatar')[0].alt).join('\n');
 osiset
+joelvh
 darrynten
 amosmos
 jedimdan
--- a/src/ShopifyApp/Http/Middleware/AuthProxy.php
+++ b/src/ShopifyApp/Http/Middleware/AuthProxy.php
@@ -6,6 +6,7 @@ use Closure;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Response;
 use function Osiset\ShopifyApp\createHmac;
+use function Osiset\ShopifyApp\parseQueryString;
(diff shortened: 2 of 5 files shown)


osiset:
Thanks!

joelvh:
@osiset thanks for merging! We're investigating some other signature validation issues our app is running into and will report back on what those cases are.

PR #516, opened 2020-07-03
Recurring annual subscription on Shopify API 2020-07

aepnat:
Shopify adds an annual subscription to the Billing API in API ver 2020-07, this can only be done with GraphQL API. I added a migration file to add new columns `interval` to the `plans` and `charges` table to flag monthly or annual recurring.

`interval` column accept two values: **EVERY_30_DAYS** or **ANNUAL**. If the interval column is `NULL`, then the interval defaults to **EVERY_30_DAYS**.

Reference:
- [Shopify blog about Release API 2020-07](https://www.shopify.com/partners/blog/shopify-api-release-july-2020#annual-charges)
- [Shopify tutorial to create an annual subscription](https://shopify.dev/tutorials/create-an-annual-subscription)

### DIFF
--- a/src/ShopifyApp/Actions/ActivatePlan.php
+++ b/src/ShopifyApp/Actions/ActivatePlan.php
@@ -74,7 +74,7 @@ class ActivatePlan
      * @param IChargeCommand $chargeCommand           The commands for charges.
      * @param IShopCommand   $shopCommand             The commands for shops.
      *
-     * @return self
+     * @return void
      */
     public function __construct(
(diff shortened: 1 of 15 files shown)


ncpope:
I haven't tested it yet but I just wanted to chime in that this has been a much needed piece of functionality since the Shopify app store was launched.

Thanks :)

osiset:
Very nice, I must've missed this news.
Changes look great, thank you, I will add a couple test cases as well shortly. :)

PR #543, opened 2020-07-30
Adds ability to delete unused script tags

niveshsaharan:
- Adds function to delete script tags
- Automatically delete unused script tags from shop

### DIFF
--- a/src/ShopifyApp/Actions/CreateScripts.php
+++ b/src/ShopifyApp/Actions/CreateScripts.php
@@ -72,6 +72,9 @@ class CreateScripts
 
         // Keep track of whats created
         $created = [];
+        $deleted = [];
+
+        $usedScriptTags = [];
         foreach ($configScripts as $scripttag) {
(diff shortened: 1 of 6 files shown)


osiset:
Thanks for this! Looks fine. Just to confirm though, memory is rusty, the API call only has access to the apps current script tags correct? I wouldn't want it to delete some other app's scripttags.

osiset:
Thanks, merging.

osiset:
@niveshsaharan This has been an ask for webhooks as well, do you think a similar process for webhooks would be beneficial?

niveshsaharan:
@osiset Maybe but I never needed to do this for webhook as we don't normally need our apps to remove existing webhooks.

osiset:
I may duplicate your process and implement this for webhooks as well.

niveshsaharan:
Sounds good. Thank you for the awesome work you're doing.

PR #546, opened 2020-07-30
Fixes #544

### LINKED ISSUE #544: Typo in table name for AddIntervalColumnToChargesTable migration
## Expected Behavior

It should remove interval column from charges table when migrations are refreshed using `php artisan migrate:refresh`

## Current Behavior

It throws error because the column is in charges table and not in plans table

niveshsaharan:
Fixes table name typo for down method in migration AddIntervalColumnToChargesTable


osiset:
Thanks!

PR #557, opened 2020-08-28
Add a .styleci.yml file

lucasmichot:
Add a `.styleci.yml` file.

@osiset, this repository needs to be activated at https://styleci.io (free for public repositories).


osiset:
I previously had this done a while ago.. not sure what happened to it. But thanks! Added and registered.

PR #558, opened 2020-08-28
Always prefer stricter comparisons

lucasmichot:
Always prefer stricter comparisons


osiset:
Interesting, must've missed this in a previous PR. Thank you.

PR #559, opened 2020-08-28
Declare missing tests classes properties

lucasmichot:
Declare missing tests classes properties, and also add their docblocks if they are missing.


osiset:
Thanks!

PR #569, opened 2020-09-09
Adds Laravel 8 support

### LINKED ISSUE #570: Add Support for Laravel 8
I was trying to add this in #569.

But dev dependency `orchestra/database` is causing problem. Since it requires Laravel 5.8
Do we need to support Laravel 5.8 in next release?

awebartisan:
Fixes #570


composer.json
```diff
@@ -17,17 +17,17 @@
         }
     ],
     "require": {
-        "php": ">=7.2",
-        "laravel/framework": "~5.8 | ^6.0 | ^7.0",
+        "php": ">=7.3",
```
Gummibeer, line 20: Minimum PHP Version of the package itself hasn't changed!?

.github/workflows/ci.yml
```diff
@@ -9,15 +9,15 @@ jobs:
     strategy:
       max-parallel: 3
       matrix:
-        php-versions: ['7.4']
-        laravel-versions: ['5.8', '6.0', '7.0']
+        php-versions: ['7.3', '7.4']
```
Gummibeer, line 12: PHP7.2 should be added to the Testmatrix.

Gummibeer (changes requested):
On top the matrix should include a prefer stable/lowest config and include the correct testbench version and exclude L5.8 on PHP7.4.

An example config is:
https://github.com/Astrotomic/laravel-translatable/blob/70811609331f10e3a068970632de1705133baa22/.github/workflows/run-tests.yml#L16

awebartisan:
Need to wait until `orchestra/database` dependency support `illuminate/support ^8.0`.

Gummibeer:
There's already a dev version of v6.x
https://packagist.org/packages/orchestra/database
So a `minimum-stability: dev` in this package composer.json should solve it.

Gummibeer (approved):

osiset:
Thanks guys! This is very much appreciated!

PR #578, opened 2020-09-12
Restore auto register service provider in composer.json

parkourben99:


osiset:
@parkourben99 The Facade no longer exists, can you update the PR to remove that part :)

PR #580, opened 2020-09-12
Migrate PHPUnit configuration file against its new schema.

lucasmichot:
Migrate PHPUnit configuration file against its new schema.


osiset:
Noticed this as well when I was running local tests, thanks for this.

osiset (approved):

Issue #596, opened 2020-09-28
Add route domain config option


squatto:
My app routes are registered like this:

```php
Route::domain(env('SHOPIFY_APP_DOMAIN'))
     ->as('app.')
     ->middleware('web')
     ->group(base_path('routes/shopify-app.php'));
```

In particular, my app's routes are served from a separate domain that is being served by the same codebase. However, the routes in this package are all being registered outside of that domain (for obvious reasons). Could an optional config option be added that defines the domain for the routes?

Another option would be to add a config option `manual_routes` that - similar to `manual_migrations` - allows us to define the routes ourselves.

I would be happy to PR either of these changes.

Thanks!

osiset:
This has been brought up before in another issue. I am adding a config to ignore internal routes.

osiset:
Closing for older ticket #595

PR #598, opened 2020-10-07
Ability to override internal routes #596

osiset:
Relates to #596, needs physical testing.

Allows for someone to list out routes they wish to be excluded from registering with Laravel.

Example: `SHOPIFY_MANUAL_ROUTES=home,billing` would tell this package to not register the internal `home` and `billing` routes.


osiset:
Physical testing passed. Merging.

Issue #599, opened 2020-10-09
Auth Shopify Middleware are not working with headers


sergeyantonuk:
## Expected Behavior

I am developing API where I use osiset package.
And my client send auth data(hmac, session, shop, timestamp, locale) through headers.
After Auth Shopify Middleware accepted the data - authentication should be successful.

## Current Behavior

We have getData() method in the vendor/osiset/laravel-shopify/src/ShopifyApp/Http/Middleware/AuthShopify.php
This method should return Array something like this:
```
array(5) {
    ["hmac"]=>
    string(64) "abcdeg3036d864177269187dd84a5d91328d8e6w99d84958"
    ["locale"]=>
    string(2) "en"
    ["session"]=>
    string(64) "abcdef273e91280396597c13e8fc2edef19d48b5efbeaa5a802"
    ["shop"]=>
    string(28) "mystore.myshopify.com"
    ["timestamp"]=>
    string(10) "1602157989"
}
```

All elements of this array are required.

So according to getData() method we have 3 possible methods to grab the Auth data:
- GET/POST parameters
- Headers
- Referer

In my case I want to use headers, but in the header's branch of code the session header is missing:

    
              ` $shop = $request->header('X-Shop-Domain');
                $signature = $request->header('X-Shop-Signature');
                $timestamp = $request->header('X-Shop-Time');
              

                $verify = [
                    'shop'      => $shop,
                    'hmac'      => $signature,
                    'timestamp' => $timestamp,  
                ];  
`
So, current behavior:
- my client send auth data(hmac, session, shop, timestamp, locale) through headers.
- Auth Shopify Middleware accepted the data and return login page(unauthenticated)
This happens because session header is missing in the getData() method

### Context

Package Version: v12.0.1
Laravel Version: v7.28.4
PHP Version: v7.2.24

darrynten:
See #601 - now you can use the JWT from Shopify directly

sergeyantonuk:
Great!
Can you tell me how can I setup it?

darrynten:
@sergeyantonuk 

The implementation can be used exactly as documented by Shopify. The backend is taken care of by this package, you must [follow the tutorial for the frontend part](https://shopify.dev/tutorials/authenticate-your-app-using-session-tokens#frontend-changes) or use a package that takes care of this for you.

[...]

Issue #588, opened 2020-09-21
Unable to verify signature.


JDSHINDE:
Hi, there
     I stuck in this error. I have app in shopify store that throwing this error in every hour, I don't know  how to resolve this, I had try lot of time but still laravel log spit this error every error.

[2020-09-21 11:56:40] production.ERROR: Unable to verify signature. {"exception":"[object] (OhMyBrew\\ShopifyApp\\Exceptions\\SignatureVerificationException(code: 0): Unable to verify signature. at /home/Myprojectname/foldername/public_html/vendor/ohmybrew/laravel-shopify/src/ShopifyApp/Middleware/AuthShop.php:171)
[stacktrace]
#0 /home/Myprojectname/foldername/public_html/vendor/ohmybrew/laravel-shopify/src/ShopifyApp/Middleware/AuthShop.php(108): OhMyBrew\\ShopifyApp\\Middleware\\AuthShop->getQueryDomain(Object(Illuminate\\Http\\Request))
#1 /home/Myprojectname/foldername/public_html/vendor/ohmybrew/laravel-shopify/src/ShopifyApp/Middleware/AuthShop.php(53): OhMyBrew\\ShopifyApp\\Middleware\\AuthShop->getShopDomain(Object(Illuminate\\Http\\Request), Object(OhMyBrew\\ShopifyApp\\Services\\ShopSession))
#2 /home/Myprojectname/foldername/public_html/vendor/ohmybrew/laravel-shopify/src/ShopifyApp/Middleware/AuthShop.php(32): OhMyBrew\\ShopifyApp\\Middleware\\AuthShop->validateShop(Object(Illuminate\\Http\\Request))
#3 /home/Myprojectname/foldername/public_html/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(171): OhMyBrew\\ShopifyApp\\Middleware\\AuthShop->handle(Object(Illuminate\\Http\\Request), Object(Closure))
#4 /home/Myprojectname/foldername/public_html/vendor/laravel/framework/src/Illuminate/Routing/Middleware/SubstituteBindings.php(41): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}(Object(Illuminate\\Http\\Request))
[...]

Issue #526, opened 2020-07-12
Auth middleware for API


sergeyantonuk:
How can I change authentification behavior? 
I am developing API where I use osiset package.
I need api to return unauthorized instead HTML form.
Is there a way to do it with this package? Or any recommendations?

As I can see we have: /vendor/osiset/laravel-shopify/src/ShopifyApp/Http/Middleware/AuthShopify.php
where we have private: handleBadVerification() method, where we have redirect to login page: `return Redirect::route('login');`

In case there is no better ways, I have a backup plan to extend AuthShopify and override handleBadVerification method, but it's private and I can't do that

osiset:
You're asking if theres an ability to instead of redirection to /login, to go somewhere else?

sergeyantonuk:
I need it for the API, so it has to return 401 Unauthorized status instead redirect to login page

osiset:
I am in the works of working on a method for this right now as login method is deprecated on Shopify.

I will include this with next update, which will throw an exception which you can catch in Laravel.

darrynten:
@sergeyantonuk can you see if the new JWT support resolves your problem?

Issue #522, opened 2020-07-07
Browser Issues: ITP, Cookies, Redirection


osiset:
**Feature**: Currently there is no support for ITP cookies, theres also lingering issues with cookies outside of my control sadly.

I am not sure if there will be a clean solution for this as adding ITP support really bloats the flow and quite honestly is frustrating to deal with the baggage Chrome and Chrome-like browsers bring with their isolated standards.

I am not sure when I can complete this, or what it will take, or even if I can do it alone. However, theres been a growing number of cookie issues (again, outside of my control), so I will attempt this as the next feature!

Any opened cookie-related issues will be closed and this issue will be referenced. I will pool your reports and use them to attempt a fix.

Thanks!

osiset:
@Kyon147 FYI. If you see any issues which may regard cookies, feel free to close and reference this issue as master :) going to pool the reports and find a way to fix ITP.

Kyon147:
> @Kyon147 FYI. If you see any issues which may regard cookies, feel free to close and reference this issue as master :) going to pool the reports and find a way to fix ITP.

No problem, also if you need some help with the issue - I'll pull a fork of the repo and try and take a look as well. As you say you are one person, so the more support the better for you and the package as a whole. If you'd like me to :)

osiset:
Any help is welcome. I had a branch last year where I did some minor ITP work, so I may revive this and massage it into the current code base and see where I can get with safari.

Kyon147:
Okay, I'll wait a little bit to see what you do with the existing work
before jumping into it.

That way don't risk doing the same stuff etc

osiset:
Following up, as a current plan of attack.

I had previous code started to help deal with ITP which showed promise. The code is no longer compatible however, but I will pick out the bits.

I will be attempting to debug this issue slowly with Safari/Chrome and try and implement ITP support as I go.

osiset:
Is anyone aware of a way to replicate this without Safari?

Firefox and Chrome I have both set to block third party cookies, but its still allowing me through.

darrynten:
We've started getting 419s on some calls (but not others) - not sure if it's related though and not seeing anything in the logs.

Am investigating, will advise.

faisal-saeed-brainx:
Hi @osiset, I am getting the samesite cookie issue in chrome incognito.

pagameba:
[...]

PR #601, opened 2020-10-12
JWT Support

### LINKED ISSUE #551: Support for new JWT auth flow
Are there anyone working on support for the new auth flow using JWT instead of cookies?

Ref:  https://shopify.dev/tools/app-bridge/authentication

darrynten:
# Shopify JWT Authentication

Provides new auth.token middleware you can place on API endpoints.

See: https://shopify.dev/tutorials/authenticate-your-app-using-session-tokens

This means you can now build apps without cookies (see #522) and will no longer experience bugs like #443. It also allows for functionality requested for in #526 and #599  because Shopify does most of the hard work now, meaning issues like #588 (and everything else cookie related) is a thing of the past.

## Usage

This package only covers the [backend functionality](https://shopify.dev/tutorials/authenticate-your-app-using-session-tokens#backend-changes) changes.

I'm not submitting any frontend changes (creating an external package instead), so follow the [frontend instructions](https://shopify.dev/tutorials/authenticate-your-app-using-session-tokens#frontend-changes) in the Shopify documentation.

You can install some Vue-based frontend scaffolding to test this with: [UnicornGlobal/laravel-shopify-vue](https://github.com/UnicornGlobal/laravel-shopify-vue).

Send the JWT you get from Shopify in the `Authorization: bearer <xyz>` header directly to any API endpoint secured with the `auth.token` middleware.

Don't forget to get a fresh JWT from Shopify on your frontend every 30 seconds so your token stays valid.

## Changes

- New Middleware for Shopify Auth Tokens `auth.token`
- API Endpoints - current user details on `/api/me` and plan list on `/api/plans`
- Split API and Web routes files
- Added missing API Key and API Secret env variables to phpunit example
- Updated HMACs in unit tests to work with default phpunit example env variables
- Bugfix for missing client_id in redirections

The reason the phpunit and HMACs have changed is due to a bug that allowed the application to operate even when there was no Shopify API key set. This PR adds a key and secret to the phpunit.dist.xml file.

**You must update your phpunit.xml file to include the missing env vars**

Also fixes an issue where the billing screen sometimes didn't know which shop to load.

Resolves #551


osiset:
Fantastic @darrynten ! Definitely would help out people! I think a small wiki addition on how to set this up would be great after too.

Tests are passing which is awesome, thanks. I was unaware of the billing screen issue, but passing in the input vars as you've done seems like a fine solution.

From my end, stuff looks fine. I'm wondering for the API error responses though, should we just throw an exception? This way it can be handled alternatively by the user if they wish to with a custom render function.

Something like: `throw new HttpException(Response::BAD_REQUEST, 'Invalid token')` for the errors?

darrynten:
@osiset I've updated the PR to throw exceptions instead of returning plain responses, and ensure that JSON responses are returned if the `accept` header is set to `application/json`.

All tests have been updated and additional tests have been added.

osiset:
@darrynten Thanks, I'll merge it!

osiset:
I'll look at tryin to add a wiki for this later the week.

jithinjoyj4:
does JWT support fix this issue #614. should I update anything, my package version is 13.1.

kouloughli-hemza:
im just confused about the database part , when first opening the app how to deal with installation process ? 
do we use the API routes ? or the package process using web , and we use JWT only for installed apps?
im way confused  about this point . 
thank you

andreuka:
Trying to understand how to make it work with server side rendered apps.

andreuka:
seems like https://github.com/turbolinks/turbolinks is the way to go

bilfeldt:
Maybe we can all assist on documenting this awesome feature :)

I am not 100% sure I get how we switch to using JWT as authentication mechanisms in a new or existing project nor when exactly it is recommended or even possible.

As far as I can see the relevant places to document this would be:
- https://github.com/osiset/laravel-shopify/wiki/Shop-Installation-and-Authentication-Process
- Maybe also a more general note about authentication and the possible setup here: https://github.com/osiset/laravel-shopify/wiki/Installation#configuration

Unresolved question that I think is worth documenting:

#### Authentication in general
There are two aspects of authentication when developing Shopify apps:
1. A user that is logged into a Shopify store needs to authenticate when interacting with a Shopify App
2. A Shopify app needs to authenticate to Shopify whenever it makes RestAPI or GraphQL requests

When installing a new app the users is goes through a rather cumbersome OAuth process as describe by Shopify [here](https://shopify.dev/tutorials/authenticate-with-oauth) and the documentation of this package [here](https://github.com/osiset/laravel-shopify/wiki/Shop-Installation-and-Authentication-Process). Essentially using the `AuthShopify` middleware on a route would ensure that a user is authenticated using OAuth correct?

Once the app is installed correctly using OAuth then a Shopify Api Token will be saved in the `password` field of the `users` table. The scope (abilities) of this Api Token is determined in the configuration of this package and determine what information the app can request from Shopify.

Each time the App makes a RestAPI or GraphQL request to Shopify then it uses this API Token to authenticate.

Now that the app is installed and the user has granted the app access to their Shopify account by going through the OAuth flow, then we need to know how to authenticate on every day use. There are mainly four interesting parts to take into account:
- User clicks on the app in the Shopify app listing
- User types the URL of the app in the browser directly
- User clicks a Shopify Admin Link which points to a route on the app
- Shopify fires a webhook

**NOTE:** How the result should look depends on wether or not the app is an [Embedded App](https://shopify.dev/concepts/apps#embedded-apps) or not.

Now for the four interesting usecases we can authenticate them differently:
- Webhooks: Webhooks contains a `X-Shopify-Hmac-SHA256 header` which can be verified using the apps shared secret - read more [here](https://shopify.dev/tutorials/manage-webhooks#verifying-webhooks)
- Admin links: I have no idea how best to authenticate these????
- Page views:  Can use OAuth which causes problems for embedded apps due to new restrictions of 3rd party cookies. An alternative is using JWT Tokens, but this seems to only work properly for Single Page Applications??

@andreuka did you manage to find out how to use this properly on a service side rendered app (most obvious example would be a multi-page app using blade files)?

Do anyone knows how this works with admin links?

andreuka:
@bilfeldt  Im using turbolinks to work with it. its not perfect and creating additional redirect when user visits first page of app. but it works.

` Route::get('/', 'Admin\Settings@set_auth')->middleware(['auth.shopify'])->name('home'); `


```
class Settings extends Controller{
    public function set_auth(){
        $shop          = Auth::user();
        $shop_id       = $shop->id;

        return view('admin.auth', [
            'shop'          => $shop,
        ]);
    }
}
```

```
@extends('layouts.default')

@section('scripts')
    @parent
@endsection
@section('content')
    <div data-link="{{ $url ?? '' }}" id="redir-to">Redirection...</div>
@endsection
```


```
var app = createApp({
                apiKey: '{{ config('shopify-app.api_key') }}',
                shopOrigin: shopOrigin,
                forceRedirect: true,
            });

            var Intervals = {};
            var redirectStarted = false;

            function redirectIfNeeded() {
                if( redirectStarted ) return false;

                if (sessionToken !== '') {
                    var redirTo = $('#redir-to');

                    if (location.pathname === '/') {
                        redirectStarted = true;
                        Turbolinks.visit("/home");
                    } else if (redirTo.length > 0) {
                        redirectStarted = true;
                        var link = redirTo.data('link');
                        Turbolinks.visit(link);
                    }
                }
            }

            function updateSessionToken(){
                getSessionToken(app).then(function(_sessionToken){
                    sessionToken = _sessionToken;

                    localStorage.setItem('sessionToken', sessionToken);

                    redirectIfNeeded();

                    $.ajaxSetup({
                        headers: {
                            'Authorization': "Bearer " + sessionToken
                        }
                    });
                });
            }


            document.addEventListener("turbolinks:load", function () {
                redirectIfNeeded();
            });


            updateSessionToken();
            setInterval(function(){
                updateSessionToken();
            }, 10000);

```

LHongy:
@andreuka Hello, I am having a hard time to understand your solutoin.
I thought we need an unauthenticated page to retrieve the session token, and then redirect to our authenticated page using that token?

osiset:
@darrynten Would you have a couple jot notes about implementation for this, off the top of your head, that I can take and turn into a wiki entry?

diemah77:
> @andreuka Hello, I am having a hard time to understand your solution, why does the route has to middleware with auth.shopify? I thought we need an unauthenticated page to retrieve the session token, and then redirect to our authenticated page using that token?

Because you need to check whether the request to your app is authorized, e.g. if the app is loaded for the first time during the installation process.

andreuka:
> > @andreuka Hello, I am having a hard time to understand your solution, why does the route has to middleware with auth.shopify? I thought we need an unauthenticated page to retrieve the session token, and then redirect to our authenticated page using that token?
> 
> Because you need to check whether the request to your app is authorized, e.g. if the app is loaded for the first time during the installation process.

Yeah, but with latest updates its not relevant anymore as there is better way.

PR #620, opened 2020-11-20
Pass session into api_init function

osiset:
Reasoning behind this is to allow for using "Custom Apps" with the same code base. One could add a migration to the users table to store the custom app api key/secret for each shop, then in the config, utilize api_init option to pull the keys from the database for the shop.

This has been requested many times since Shopify now limits how apps work, you can not have an unpublished app installed on many stores. This "solution" will allow for same code base, and switching the API keys on the fly when the API interface is requested for the shop.

### DIFF
--- a/src/ShopifyApp/Services/ApiHelper.php
+++ b/src/ShopifyApp/Services/ApiHelper.php
@@ -49,7 +49,11 @@ class ApiHelper implements IApiHelper
         // Create the instance
         if ($this->getConfig('api_init')) {
             // User-defined init function
-            $this->api = call_user_func($this->getConfig('api_init'), $opts);
+            $this->api = call_user_func(
+                $this->getConfig('api_init'),
+                $opts,
+                $session
+            );
         } else {
             // Default init
             $ts = $this->getConfig('api_time_store');
--- a/src/ShopifyApp/resources/config/shopify-app.php
+++ b/src/ShopifyApp/resources/config/shopify-app.php
@@ -233,7 +233,8 @@ return [
     |--------------------------------------------------------------------------
     |
     | This option is for initing the BasicShopifyAPI package optionally yourself.
-    | The first param injected in the current options (\Osiset\BasicShopifyAPI\Options).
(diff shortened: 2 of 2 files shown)


osiset:
Example code for `config/shopify-app.php`:

```php
    'api_init' => function (Options $opts, ?Session $session) {
        $ts = Config::get('shopify-app.api_time_store');
        $ls = Config::get('shopify-app.api_limit_store');
        $sd = Config::get('shopify-app.api_deferrer');

        if ($session) {
            $shop = User::where('name', $session->getShop())->first();

            $opts->setApiKey($shop->api_key);
            $opts->setApiSecret($shop->api_secret);
        }

        return new BasicShopifyAPI(
            $opts,
            new $ts(),
            new $ls(),
            new $sd()
        );
    },
```

Example migration:

```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

class AddIntervalColumnToPlansTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::table('plans', function (Blueprint $table) {
            $table->string('interval')->nullable()->after('price');
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::table('plans', function (Blueprint $table) {
            $table->dropColumn('interval');
        });
    }
}
```

osiset:
Similarly but with ENV instead of a database lookup.

```php
    'api_init' => function (Options $opts, ?Session $session) {
        $ts = Config::get('shopify-app.api_time_store');
        $ls = Config::get('shopify-app.api_limit_store');
        $sd = Config::get('shopify-app.api_deferrer');

        if ($session) {
            $shopDomain = strtoupper(explode('.', $session->getShop())[0]);
            $opts->setApiKey(env("API_KEY_{$shopDomain}"));
            $opts->setApiSecret(env("API_SECRET_{$shopDomain}"));
        }

        return new BasicShopifyAPI(
            $opts,
            new $ts(),
            new $ls(),
            new $sd()
        );
    },
```

PR #635, opened 2020-12-08
Test for PHP 8

lucasmichot:
Test for PHP 8

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ jobs:
           - '7.2'
           - '7.3'
           - '7.4'
+          - '8.0'
         laravel:
             - '5.8'
             - '6.0'
@@ -62,7 +63,7 @@ jobs:
       run: vendor/bin/phpunit
 
     - name: Upload coverage results
-      if: matrix.php == '7.4' && matrix.laravel == '8.0'
+      if: matrix.php == '8.0' && matrix.laravel == '8.0'
       env:
         COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |


osiset:
I ran into an issue where the value object package is linked to version 7. Not sure if its active anymore, may require to be forked.

osiset:
Digging more into this.

php-array-of package is locked to ^7.1, composer file would need to be updated to be ^7.1 || ^8.0

That package is used by valueobjects, which is also locked to ^.7.1.

So both packages would need to be updated.

I've done the work for php-array-of and submitted a PR. I'm not sure if the maintainer is active anymore, so if it sits for a bit and is not accepted, I can absorb both packages I guess would be quickest.

lucasmichot:
Hey @osiset , yes the dependency is also locked.
the package has not been updated for years. I feel this is somehow safer to switch to another dependency?
WDYT ?

osiset:
@lucasmichot I'll open the PR today, work is done. I'll leave it for a few days to see if it gets accepted, if not, I'll finish the work and release it to packagist under this namespace and then you can swap the dependency in composer.json to the new namespace and all should work well then.

osiset:
@lucasmichot PR was accepted for php-array-of, so maintainer is still active :) once patch is on packagist, I'll give the maintainer a PR for the other package... then we should be good to go.

osiset:
@lucasmichot Both PRs done, update valueobjects package in composer.json to 0.5 and it'll resolve.

I think Laravel 5.8 needs to be excluded from the matrix for PHP 8.0 if I remember as well.

NESTEROVILYASERGEEVICH:
Hey @osiset,
I am using your package, I need your help with some installation.
Could you please help me?

osiset:
@lucasmichot PHP 8 support is in master now.

Things I did:

- 8.22 is the lowest Laravel which supports PHP 8 and doesn't have reported security issues
- Dropped Laravel 5 and 6
- Modified action to exclude PHP 7.2 with Laravel 8.22 (since Laravel 8.22 wants PHP 7.3 and up)
- Modified action to exclude PHP 8.0 with Laravel 7.0 (since Laravel 7.0 wants PHP 7.2)

osiset:
Ill close this PR for now since PHP8 support was released. Feel free to push any improvements to it however that you see fit :D

PR #681, opened 2021-02-04
Fix make webhook stub

savchukoleksii:
Currently webhook stub use interface:
`use Osiset\ShopifyApp\Contracts\Objects\Values\ShopDomain;`

So this code produce error:
`public function handle()
{
    // Convert domain
    $this->shopDomain = ShopDomain::fromNative($this->domain);

    // Do what you wish with the data
    // Access domain name as $this->shopDomain->toNative()
}`

Right one use statement `use Osiset\ShopifyApp\Objects\Values\ShopDomain;`


naoufalb99:
It's should be 
```$this->shopDomain = ShopDomain::fromNative($this->shopDomain);```

Because ```$this->domain``` doesn't exist

PR #722, opened 2021-03-05
Added read shop from referrer if everything fails

andthink:
when using config_api_callback in configuration sometimes $shop is just null if i click on an internal url of my app, but the information is present in the referrer

### DIFF
--- a/src/ShopifyApp/Services/ApiHelper.php
+++ b/src/ShopifyApp/Services/ApiHelper.php
@@ -475,7 +475,7 @@ class ApiHelper implements IApiHelper
      * @param  Session  $session
      * @return mixed
      */
-    private function getShop(Session $session)
+    private function getShop(Session $session = null)
     {
         if ($session) {
             return $session->getShop();
--- a/src/ShopifyApp/Services/ApiHelper.php
+++ b/src/ShopifyApp/Services/ApiHelper.php
@@ -45,7 +45,7 @@ class ApiHelper implements IApiHelper
         parse_str(Request::server('HTTP_REFERER'), $refererQueryParams);
         $refererShop = $refererQueryParams['shop'] ?? null;
 
-        $shop = $session ? $session->getShop() : Arr::get(Request::all(), 'shop') ?? $shop;
+        $shop = $session ? $session->getShop() : Arr::get(Request::all(), 'shop') ?? $refererShop;
         $opts->setApiKey(getShopifyConfig('api_key', $shop));
         $opts->setApiSecret(getShopifyConfig('api_secret', $shop));
         $opts->setVersion(getShopifyConfig('api_version', $shop));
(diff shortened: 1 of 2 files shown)


osiset:
Thanks! Given this situation can happen with other cases, maybe you can include support for headers too?

If you look at AuthShopify.php::getData, you'll see code that grabs from input (shop query param), headers (x-shop-domain), and referrer.

andthink:
something like this?
` $requestShop = Arr::get(Request::all(), 'shop');
        parse_str(Request::server('HTTP_REFERER'), $refererQueryParams);
        $refererShop = Arr::get($refererQueryParams,'shop');
        $headerShop = Request::header('X-Shop-Domain');

        $shop = $session ? $session->getShop() : $requestShop  ?? $refererShop ?? $headerShop;`

osiset:
The multiple nullcoalescing is a bit hard on the eyes.

My suggestion would be to extract the shop detection into its own private function and have the function return the shop to use.

andthink:
> The multiple nullcoalescing is a bit hard on the eyes.

ahahah ok

> 
> My suggestion would be to extract the shop detection into its own private function and have the function return the shop to use.

ok I'll look into it

src/ShopifyApp/resources/views/layouts/default.blade.php
```diff
@@ -34,4 +34,4 @@
 
         @yield('scripts')
     </body>
-</html>
\ No newline at end of file
+</html>
```
andthink, line 37: This was added without me noticing, but isn't it needed?

andthink:
Ok, I've fixed automatic test errors and added also a parameter to bridge createApp function, without that config_api_callback did not work.
I was working before on version 14 and didn't have this problem, now I've updated and i needed to fix also  this

`apiKey: '{{ \Osiset\ShopifyApp\getShopifyConfig('api_key', Auth::user()->name ) }}',`

src/ShopifyApp/resources/views/layouts/default.blade.php
osiset: Thats fine!

osiset:
Ill review now thanks!

src/ShopifyApp/Services/ApiHelper.php
```diff
@@ -470,4 +470,29 @@ protected function doRequestGraphQL(string $query, array $payload = null)
 
         return $response;
     }
+
+    /**
+     * @param  Session  $session
+     * @return mixed
+     */
+    private function getShop(Session $session = null)
```
osiset, line 478: Lets have this function return `ShopDomain` value to ensure the shop domain is formatted correctly.
Then line 45, we can do `->toNative()` to have it back into a string for use with configs.

andthink:
@osiset Done! I still have cookie/session issues in some cases, but I manage with the ?shop= parameter for now.

andthink:
Found a breaking bug in ShopDomain::fromNative use, please @osiset wait for next UPDATE

osiset:
Sure :)

darrynten:
this is shockingly dangerous 😱

PR #730, opened 2021-03-09
Use named routes from config

squatto:
There were a few places that weren't using named routes from the config, so I tracked down all that I could find and fixed them.


osiset:
Thank you!

squatto:
You're welcome!

PR #732, opened 2021-03-10
bugfix/593

osiset:
* Ignore data sources on AuthShopify middleware when HMAC is not present


src/ShopifyApp/Http/Middleware/AuthShopify.php
```diff
@@ -230,6 +228,51 @@ private function getHmac(Request $request): ?array
...
+                return $refererQueryParams['shop'];
+            },
+        ];
+
+        // Loop through each until we find the HMAC
+        foreach ($options as $method => $value) {
+            $result = is_callable($value) ? $value() : $value;
+            if ($result !== null) {
+                // Found a shop
+                return ShopDomain::fromNative($result);
```
darrynten, line 268: won't this just return whatever shop i inject into the http_referer variable?

PR #780, opened 2021-05-07
Feature/cookieless - changes for turbolinks, install app 

Enmaboya:
In my case, when I tried to install the application, I was immediately redirected to /authenticate/token, which ended in an error. That's why I added a store check.

I also changed a little receiving and updating the token before sending Turbolinks request.
I also removed const from variables, because Turbo conflicts with constant variables.

Maybe it would be more correct to put them on a separate template

### DIFF
--- a/src/ShopifyApp/Http/Middleware/VerifyShopify.php
+++ b/src/ShopifyApp/Http/Middleware/VerifyShopify.php
@@ -115,8 +115,14 @@ class VerifyShopify
 
         $tokenSource = $this->getAccessTokenFromRequest($request);
         if ($tokenSource === null) {
-            // Not available, we need to get one
-            return $this->handleMissingToken($request);
+            //Check if there is a store record in the database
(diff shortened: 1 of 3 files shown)


osiset:
@Enmaboya Interesting, I didn't see an issue for install, but good to note.

> I also removed const from variables, because Turbo conflicts with constant variables.

Can you explain why this is?

Additionally, you're saying with this PR, install flow is working, and once installed, things are OK as well? I'll need to adjust my tests as well.

Enmaboya:
> Can you explain why this is?

turbolinks evaluates <script> elements in a page’s <body> each time it renders the page.

but I was a little hasty and returned the constants.
now turbolinks will just ignore "app bridge"

> install flow is working, and once installed, things are OK as well?

Yes, the installation completes correctly with this method.

Without it, VerifyShopify middleware tries to get the token ($tokenSource = $this->getAccessTokenFromRequest($request)), but since it's not there during installation, it redirects me to "/authenticate/token"

PR #784, opened 2021-05-10
Find shop domain in request when getting the token

squatto:
When redirecting within your app, it's easy to get into a state where there isn't a "shop" query param for the token controller to use. At that point you're completely stuck because you can't auth anymore. You're at the end of the road.

This PR adds a static method `ShopDomain::getFromRequest()` that tries to find the token in the request in multiple ways:

1. If the request inputs contain "shop" or "shopDomain", it is used.
2. If the request headers contain "X-Shop-Domain", it is used.
3. If the request headers have a referrer URL then the same checks are performed against it ("shop" and "shopDomain")
4. If the referrer URL contains a "token" query param, it is decoded and checked for a shop domain.
    - To accomplish this, I added an optional param to the `SessionToken` constructor that allows us to *not* verify the token. This allows us to decode it and get the values regardless of its validity/expiration. It defaults to verifying the token, which is the original functionality.

The end result is that in most cases, the token controller is able to find the shop and you can continue forward as expected.

### DIFF
--- a/src/ShopifyApp/Objects/Values/ShopDomain.php
+++ b/src/ShopifyApp/Objects/Values/ShopDomain.php
@@ -5,8 +5,9 @@ namespace Osiset\ShopifyApp\Objects\Values;
 use Assert\AssertionFailedException;
 use Funeralzone\ValueObjects\Scalars\StringTrait;
 use Illuminate\Http\Request;
+use Illuminate\Support\Arr;
 use Osiset\ShopifyApp\Contracts\Objects\Values\ShopDomain as ShopDomainValue;
-use Osiset\ShopifyApp\Exceptions\MissingShopDomainException;
+use Osiset\ShopifyApp\Objects\Enums\DataSource;
 use function Osiset\ShopifyApp\getShopifyConfig;
 use function Osiset\ShopifyApp\parseQueryString;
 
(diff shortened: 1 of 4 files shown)


squatto:
Ah, actually, I just found `VerifyShopify::getShopDomainFromRequest()`. I'll modify that slightly and use it.

osiset:
@squatto Brain was slow this morning. Since the base value object implementation uses a static method of ::fromNative, like `ShopDomain::fromNative($request->get('shop'))`, I'm thinking the `getFromRequest` could be changed to `fromRequest(Request $request)`, for some consistency?

squatto:
> @squatto Brain was slow this morning. Since the base value object implementation uses a static method of ::fromNative, like `ShopDomain::fromNative($request->get('shop'))`, I'm thinking the `getFromRequest` could be changed to `fromRequest(Request $request)`, for some consistency?

I'm good with that. I think it's definitely more consistent. Do you want me to refactor and submit a PR?

osiset:
@squatto I can grab it and do it for you :) no worries.

src/ShopifyApp/Objects/Values/ShopDomain.php
```diff
@@ -25,6 +30,70 @@ public function __construct(string $domain)
...
+     * @param Request $request The request object.
+     *
+     * @return ShopDomainValue
+     */
+    public static function getFromRequest(Request $request): ShopDomainValue
+    {
+        // All possible methods
+        $options = [
+            // GET/POST
+            DataSource::INPUT()->toNative() => $request->input('shop') ?? $request->input('shopDomain'),
```
darrynten, line 50: doesn't this make it possible to easily spoof any shop?

PR #790, opened 2021-05-17
Update jQuery ajax header Authorization setting

thang12l:
Correct Authorization Bearer setting for jQuery ajax header

### DIFF
--- a/src/ShopifyApp/resources/views/partials/token_handler.blade.php
+++ b/src/ShopifyApp/resources/views/partials/token_handler.blade.php
@@ -23,8 +23,10 @@
 
         if (window.jQuery) {
             // jQuery
-            window.jQuery.ajaxSettings.headers['Authorization'] = `Bearer ${window.sessionToken}`;
-        }
+            // window.jQuery.ajaxSettings.headers['Authorization'] = `Bearer ${window.sessionToken}`;
+            window.jQuery.ajaxSetup({
+                headers: { 'Authorization': 'Bearer ' + window.sessionToken}
+            });        }
 
         if (window.axios) {
             // Axios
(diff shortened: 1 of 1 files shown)


squatto:
@thang12l doesn't doing it the way that you're proposing overwrite any existing custom headers? We changed it to directly set the "Authorization" value in the `headers` array to avoid that.

thang12l:
@squatto the proposing way will overwrite custom headers if they exist. Just update it again to avoid it. When no headers exist, use ajaxSetup to add the new setting.

src/ShopifyApp/resources/views/partials/token_handler.blade.php
```diff
@@ -23,7 +23,13 @@
 
         if (window.jQuery) {
             // jQuery
-            window.jQuery.ajaxSettings.headers['Authorization'] = `Bearer ${window.sessionToken}`;
+            if (window.jQuery.ajaxSettings.headers) {
+                window.jQuery.ajaxSettings.headers['Authorization'] = `Bearer ${window.sessionToken}`;
+            } else {
+                window.jQuery.ajaxSetup({
+                    headers: { 'Authorization': 'Bearer ' + window.sessionToken}
```
osiset, line 30: @thang12l Can you adjust this line to use template syntax like above :) I can merge then.

thang12l (commented):
Just updated with the template style

osiset:
Thanks! I havent used jQuery in a long time, but if `window.jQuery.ajaxSettings.headers` is `undefined`, can we just `window.jQuery.ajaxSettings.headers = {};` ? Or does that not work, as in, you have to previously have used ajaxSetup?

thang12l:
@osiset window.jQuery.ajaxSettings.headers = {}; will work like a charm!

osiset:
@thang12l Nice! Update your PR and ill merge it in.

thang12l:
@osiset , Already done with 7a0aa37c10f4089590d9286e3cc21b7b6a43b3a6. Please merge it in.

PR #794, opened 2021-05-27
fix the bugs to redirect to login page if auth empty

codesharpdev:
Problem:
on localhost, geting error "Osiset\ShopifyApp\Exceptions\MissingShopDomainException".  

Solution 
here I have redirected to login page if shop parameter auth domain empty the login page will display, there user/developer can add their shop url to continue the installation of app.

also, I have added login.blade.php and route entry example code with steps in the installtion wiki guide

### DIFF
--- a/src/ShopifyApp/Http/Middleware/AuthShopify.php
+++ b/src/ShopifyApp/Http/Middleware/AuthShopify.php
@@ -352,8 +352,8 @@ class AuthShopify
     private function handleBadVerification(Request $request, ShopDomainValue $domain)
     {
         if ($domain->isNull()) {
-            // We have no idea of knowing who this is, this should not happen
-            throw new MissingShopDomainException();
+            //Rediect to login page if auth domain empty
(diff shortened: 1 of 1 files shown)


osiset:
There is no login route or file part of the library anymore. The throw is in there so developers can do what they wish. You can create an exception handler to do the redirect :)

ncpope:
Here is how I currently handle it in one of my apps:

in app\Exceptions\Handler

```php
    /**
     * Register the exception handling callbacks for the application.
     *
     * @return void
     */
    public function register()
    {
        // ... Some other stuff that is unrelated

        $this->renderable(function (MissingShopDomainException $e, $request) {
            return redirect()->route('login');
        });
    }
```

and then in my routes/web

```php
Route::get('/login', [LoginController::class, 'renderLogin'])->name('login');
```

Kyon147:
Closing this as it is intentional to throw the exception.

PR #810, opened 2021-06-18
GitHub actions

lucasmichot:
This PR improves and speed up CI tests.

- Normalize the YAML action file and respect `.editorconfig` settings
- Rename the job according to the Github action filename.
- Ensure the coverage test happens once every PR.
- No need to validate and normalize composer.json so for each version of PHP


osiset:
`matrix.php == '8.22'` :facepalm: ugh, how'd I miss that, oops. Thanks for the help!

PR #811, opened 2021-06-18
Also test the message and code of ALL expected exceptions.

lucasmichot:
This PR tests the message and code of ALL expected exceptions.


osiset:
Great, thanks! I debated on this internally with myself, originally I thought it would be overkill, but its nice to have, especially if a message in the future contains something dynamic (like a shopname).

PR #812, opened 2021-06-18
Ensure that all classes names are PSR4 compliant.

lucasmichot:
This PR ensures that all classes names are PSR4 compliant.


osiset:
:facepalm: Another oopsie, thanks!

PR #818, opened 2021-06-22
Only install xdebug when necessary

lucasmichot:
Only install `xdebug` when necessary


lucasmichot:
Closing after being split into various PR

PR #821, opened 2021-06-23
Replace abandoned package cedx/coveralls

lucasmichot:
Replace abandoned package `cedx/coveralls` by `php-coveralls/php-coveralls`


osiset:
Had no idea it was abandoned, thanks.

PR #827, opened 2021-06-23
Do not use env() helper outside the config file.

lucasmichot:
Do not use `env()` helper outside the config file.


osiset:
Noticed ITP in there too, I'll remove this afterwards. Thanks!

PR #834, opened 2021-06-23
Introduce some Static analysis.

lucasmichot:
Introduce some Static analysis with `phpstan`.

For now the level is only 1, the goal is to do step by step to 8


src/ShopifyApp/ShopifyAppProvider.php
```diff
@@ -103,12 +103,7 @@ public function register()
 
             // Queriers
             IShopQuery::class => [self::CSINGLETON, function () {
-                $model = $this->app['config']->get('auth.providers.users.model');
-                $modelInstance = new $model();
-
-                return new ShopQuery(
-                    $modelInstance
-                );
```
osiset, line 111: Ah! Didn't realize I still had this left in, since its internally handled in the class now.

PR #839, opened 2021-06-23
Update composer.json

lucasmichot:
Update `composer.json` :
- add new keywords (taken from Github topics)
- add `forum` entry
- add `wiki` entry


osiset:
Didn't know about the wiki and forum entries, nice.

PR #840, opened 2021-06-23
Use a Helpers class

lucasmichot:
Replace helpers with public static method contained in a `Helpers` class.


lucasmichot:
Closed by https://github.com/osiset/laravel-shopify/pull/844

PR #841, opened 2021-06-23
Use ::class notations for routes controllers

lucasmichot:
Use `::class` notations for routes controllers


osiset:
I had this in previously and had the same issues with the test suite passing. Although it would work locally. ::class was introduced in PHP5.5 I believe, so I don't know why it would have an issue.

lucasmichot:
Yeah, I think to investigate this a bit more :-D

lucasmichot:
I could make the tests pass @osiset

PR #853, opened 2021-06-30
change middleware whitelist

Enmaboya:
changing the allowed middleware for safari and incognito mode in chrome

### DIFF
--- a/src/ShopifyApp/Http/Middleware/VerifyShopify.php
+++ b/src/ShopifyApp/Http/Middleware/VerifyShopify.php
@@ -97,7 +97,7 @@ class VerifyShopify
         }
 
         // Continue if current route is an auth or billing route
-        if (Str::contains($request->getRequestUri(), ['/authenticate', '/billing'])) {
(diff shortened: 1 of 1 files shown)


osiset:
I have not had an issue with incognito. What is the issue you're experiencing?
With this change, an initial request to /billing, will put them through the token route before hitting. This is fine, just extra redirects. Just curious to how its not working if you know? We are passing the shop param to and from,

Enmaboya:
> I have not had an issue with incognito. What is the issue you're experiencing?
> With this change, an initial request to /billing, will put them through the token route before hitting. This is fine, just extra redirects. Just curious to how its not working if you know? We are passing the shop param to and from,

In my case, the app uses "freemium" mode, so when redirected to the route "billing" my $request is empty and I can not get the store.
This problem, in fact, exists only in safari 13/14 and private mode.

It is solved quite simply by adding a get parameter:

"{{ route('billing', ['plan' => $plan->id, 'shop' => auth()->user()->name]) }}">

osiset:
Interesting, I added your Blade markup to the wiki entry for billing.

PR #868, opened 2021-07-06
Add backward compatibility with REST-style webhook topics (e.g. "orders/create")

squatto:
v17 requires webhooks to be configured with GraphQL-style webhook topics (e.g. "ORDERS_CREATE") instead of the previously-used REST-style webhook topics (e.g. "orders/create"). If there are webhooks in the config and `topic` isn't changed on all of them during the upgrade process, an exception is thrown that it isn't terribly clear how to fix: `Variable $topic of type WebhookSubscriptionTopic! was provided invalid value`.

This PR adds backward compatibility with REST-style webhook topics through the use of a new utility method: `Util::getGraphQLWebhookTopic(string $topic): string`

```php
Util::getGraphQLWebhookTopic('orders/created') === 'ORDERS_CREATED'
```

When the webhooks in the config are being created on the store, the `topic` is run through this method to ensure that the GraphQL-style topic is being used, even if the topic in the config is REST style.

---

I also updated the `shopify-app:make:webhook` command output to include the GraphQL-style topic and the actual webhook handler URL (instead of a generic URL). For example:

```
$ php artisan shopify-app:make:webhook OrdersCreate orders/create
For non-GDPR webhooks, don't forget to register the webhook in config/shopify-app.php. Example:

    'webhooks' => [
        [
            'topic' => 'ORDERS_CREATE',
            'address' => 'https://shopifyapptest.ngrok.io/webhook/orders-create'
        ]
    ]
```

### DIFF
--- a/src/ShopifyApp/Console/WebhookJobMakeCommand.php
+++ b/src/ShopifyApp/Console/WebhookJobMakeCommand.php
@@ -4,6 +4,7 @@ namespace Osiset\ShopifyApp\Console;
 
 use Illuminate\Foundation\Console\JobMakeCommand;
 use Illuminate\Support\Str;
+use Osiset\ShopifyApp\Util;
 use Symfony\Component\Console\Input\InputArgument;
 
 class WebhookJobMakeCommand extends JobMakeCommand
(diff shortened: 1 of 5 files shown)


squatto:
Thanks @osiset! 👍🏻

PR #873, opened 2021-07-08
Turbulinks fixes

Enmaboya:
Sometimes when turbolinks were turned on the application did not install correctly

### DIFF
--- a/src/ShopifyApp/Traits/AuthController.php
+++ b/src/ShopifyApp/Traits/AuthController.php
@@ -27,12 +27,13 @@ trait AuthController
     public function authenticate(Request $request, AuthenticateShop $authShop)
     {
         // Get the shop domain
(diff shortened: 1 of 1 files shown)


src/ShopifyApp/Traits/AuthController.php
```diff
@@ -27,12 +27,13 @@ trait AuthController
     public function authenticate(Request $request, AuthenticateShop $authShop)
     {
         // Get the shop domain
-        if (Util::getShopifyConfig('turbo_enabled') && $request->user()) {
-            // If the user clicked on any link before load Turbo and receiving the token
-            $shopDomain = $request->user()->getDomain();
+        $shopDomain = $request->has('shop')
```
darrynten, line 30: this doesn't feel very safe

also where did the check if turbo is even enabled go?

PR #875, opened 2021-07-08
Also cache laravel/legacy-factories and illuminate/contracts.

lucasmichot:
Improve cache with taking in consideration `laravel/legacy-factories` and `illuminate/contracts`.


osiset:
Good call. Eventually we can remove this once Laravel 7 is old enough.

PR #879, opened 2021-07-08
Enforce PSR12 standards.

lucasmichot:
Enforce [PSR12](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/doc/ruleSets/PSR12.rst) standards.


osiset:
@lucasmichot Thanks, is this something that can be checked in GithubActions too?

lucasmichot:
> @lucasmichot Thanks, is this something that can be checked in GithubActions too?

This could be check with `php-cs-fixer` as there is a PSR12 rule available.
But this would also defeat a bit the goal of StyleCI (that does not come with a PSR12 rule set)

PR #889, opened 2021-07-12
Optimize conditions

lucasmichot:
This PR optimizes many conditions.


src/ShopifyApp/Services/CookieHelper.php
```diff
@@ -67,23 +67,21 @@ public function checkSameSiteNoneCompatible(): bool
...
         ) {
             $compatible = true;
         }
 
-        if ($this->agent->is('UCBrowser') &&
-            $browser['float'] > 12.13
-        ) {
+        if ($browser['float'] > 12.13 && $this->agent->is('UCBrowser')) {
             $compatible = true;
         }
```
osiset, line 87: I dont think this file is used anymore... I'll double check and remove later.

PR #897, opened 2021-07-15
Add missing return types.

lucasmichot:
Add missing return types.


osiset:
@lucasmichot Accidental close?

lucasmichot:
On purpose @osiset , but still in my plans

PR #898, opened 2021-07-15
Fix PHPStan CI arguments syntax.

lucasmichot:
This PR fixes PHPStan CI arguments syntax.


lucasmichot:
Fixed in https://github.com/osiset/laravel-shopify/pull/899

PR #900, opened 2021-07-15
:bug: Retain host parameter after successful install

stevesweets:
AppBridge 2.0 (now required for successful submission) requires the host parameter, after redirect it is not available on the URL any longer. Include it here to make it easily available.

I'm not using the supplied views, mine are custom, and I need the host parameter to be more readily available when I instantiate appbridge.  This works fine.


stevesweets:
Static analysis appears to be failing for a reason that isn't related to the code change; Want a hand with that?

osiset:
Memory error, no worries, thanks!

Issue #932, opened 2021-08-06
Some customers can't install the app unless deleted the user's info in the database


heeeydex:
## Expected Behavior

When a user installed the app, it should redirect to the billing page and can use the app normaly

## Current Behavior

Some of the users are experiencing "There's no page at this address". password is blank in the database.

## Failure Information

We can't reproduce the issue. But some of our users are experiencing this issue and we can see that there is a lot of users which don't have value in their password field. They can successfully use the app if we remove their information from our database and ask them to reinstall the app.

https://ibb.co/WDqcHgn

### Steps to Reproduce
We can't reproduce the error but a lot of our customer is facing the issue.

1. Go to https://apps.shopify.com/copify-copy-products
2. Install the app

### Context

* Package Version: ^17.0
* Laravel Version: ^8.40
* PHP Version: 7.4
* Using Digital Ocean, Polaris, Ploi.iio

### Failure Logs
Logs from Sentry.io
https://ibb.co/SBxRvgv

heeeydex:
@osiset I'm not sure if this is related to this https://github.com/osiset/laravel-shopify/issues/919

heeeydex:
Yes. I can confirm it's related to #919 I tried clicking Add App and entered my shopify store URL. Then went back to https://apps.shopify.com/copify-copy-products then click Add App again

PR #933, opened 2021-08-06
Fixed installation issue

heeeydex:
fixed ignoring installation the first time, will cause an installation failure next time for #932


jordyvanderhaegen:
Any idea on when this will be released? Shopify seems to be taking apps down due to this exact issue. We used a fork of this PR in production and it seems to be working fine 👍

PR #934, opened 2021-08-06
UPDATE: configurable auth provider

vkislichenko:
I'm sure there is a valid reasoning behind current logic of extending **User** model with shopify shop trait,
But I really do need them to be separate.

This PR provides one more shopify app config option `shop_auth_provider` ("users" by default)
Model is then taken from config `auth.providers.{$config['shop_auth_provider']}.model`

Works if manual migrations are used and user table is replaced with custom one
Models provided by this package will still reference custom model using `user_id` columns, but it seems like a minor detail related to bigger refactoring.


osiset:
Sounds good to me! Please see the failed lint, adjust, and I'll merge :)

vkislichenko:
UPDATE: added config option `shop_auth_guard` 
It's mostly to prevent use of custom default auth guards (like jwt), but it might be also useful in other cases when `shop_auth_provider` specifies a custom model

src/Http/Middleware/AuthProxy.php
```diff
@@ -82,6 +82,11 @@ public function handle(Request $request, Closure $next)
         // Login the shop
         $shop = $this->shopQuery->getByDomain($shop);
         if ($shop) {
+            // Override auth guard
+            if(($guard = Util::getShopifyConfig('shop_auth_guard'))) {
```
osiset, line 86: @vkislichenko Fails linting here, needs a space after the if

src/Http/Middleware/VerifyShopify.php
```diff
@@ -248,6 +248,11 @@ protected function loginShopFromToken(SessionToken $token, NullableSessionId $se
             return false;
         }
 
+        // Override auth guard
+        if(($guard = Util::getShopifyConfig('shop_auth_guard'))) {
```
osiset, line 252: Same as previous @vkislichenko , needs space after if to pass linting.

osiset (changes requested):
Couple small issues to correct linting, then can be merged.

osiset (approved):

osiset:
Checks passed, linting pass. Merging. Thanks @vkislichenko !

salahy:
What about the migrations ? they still change the user model

vkislichenko:
@salahy 
> What about the migrations ? they still change the user model

there's "Manual migrations" in config
and command `php artisan vendor:publish --tag=shopify-migrations` to push migrations to your app's folder so you're free to modify before migrating.

nahid:
When this changes will be released? @osiset

PR #935, opened 2021-08-07
Non embedded

xplodedthemes:


osiset:
Thanks for this.

Things I see that I will attempt to work on...

- AuthorizeShop... is removed in v17, we can somehow massage this into whats existing on v17
- Removal of CookieHelper... dont think its needed anymore
- Removal of ShopSession... v17 uses ShopContext injected into the model, somehow massage this in

osiset:
@xplodedthemes Is there anything else missing, just curious on a couple things as its been a while since I touched non-embedded.

1. Is there a setting to toggle embedded and non-embedded? Or are you simply swapping the middleware
2. More with point 1... is there Blade layout modifications to not load appbridge?
3. Whats the current flow, are you still pointing to /authenticate

aveeto:
@osiset 

That's it, I think.

1. It is all based on the **appbridge_enabled** option. 
2. Nothing has changed in the views. There was already a condition whether to load app bridge or not within the layout based on the **appbridge_enabled** option
3. Yes, I brought back to the **authenticate.oauth** route from v16.

In v16, there was a login page view that I am currently still using. We can actually use this view for both Embedded and Non-Embedded apps, and it would work fine. The only thing is that, with the non-embed, it will always re-direct to the login page since the session is not saved. It becomes more of an app install page instead of a login page.

osiset:
I think app_bridge_enabled option is currently removed in master? I'd have to check, if not, then that makes sense and the combined work can begin.

xplodedthemes:
@osiset No, the option is still there in master: **appbridge_enabled**. Note that it's appbridge without underscore.

https://github.com/osiset/laravel-shopify/blob/master/src/resources/config/shopify-app.php

valdisd:
Any updates here? Would love for this to be merged.

osiset:
@valdisd Not at this time, still trying to get time to work on this.

valdisd:
I tested this and when installing this goes into /shopify/authenticate/oauth redirect loop.

Works if the installation is done with app bridge enabled and disabled afterwards. Will try to locate the problem.

osiset:
@valdisd I believe the code provided by OP works or partially works, I just need it reworked into the existing code to not have duplication. I outlined some points previously, but feel free to poke around :)

apurbajnu:
Any update on this?

o2b3k:
@osiset  Hi any update on this! I really need this feature

t08-apporio:
hi @xplodedthemes composer installed your fork. it's currently stuck in a redirect loop. The url in question is `authenticate/oauth`. I updated `appbridge_enabled` to false and also disabled the embedded app option in app setup. Any idea what the issue could be?

@Aryanhasanzadeh were you able to make any progress?

Issue #950, opened 2021-08-18
Session in Safari browser - Issue


KevinSHansen:
*For bug reporting only! If you're posting a feature request or discussion, please ignore.*

## Expected Behavior

The app is working perfect in Chrome with session, but as soon as i open the app in Safari the user isnt authorized. 

## Current Behavior

Auth is never set in Safari and Shopify have been denying the app for this before.

## Failure Information

Using v17.0.1 and updated everything to use Sessions.

### Steps to Reproduce

Updated to v17.0.1 and followed the upgrade guide.

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Package Version: v17.0.1
* Laravel Version: v8.12
* PHP Version: v8
* Using a toolset (Docker, Laradock, Vagrant, etc.): Kubernetes (nginx)

### Failure Logs
`{
    "class": "Facade\\Ignition\\Exceptions\\ViewException",
    "message": "Attempt to read property \"name\" on null (View: /var/www/html/vendor/osiset/laravel-shopify/src/resources/views/layouts/default.blade.php)",
    "code": 0,
    "file": "/var/www/html/vendor/osiset/laravel-shopify/src/resources/views/layouts/default.blade.php:33",
    "trace": [
        "/var/www/html/vendor/osiset/laravel-shopify/src/resources/views/layouts/default.blade.php:33",
        "/var/www/html/vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php:107",
        "/var/www/html/vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php:108",
        "/var/www/html/vendor/laravel/framework/src/Illuminate/View/Engines/PhpEngine.php:58",
        "/var/www/html/vendor/livewire/livewire/src/ComponentConcerns/RendersLivewireComponents.php:35",
        "/var/www/html/vendor/laravel/framework/src/Illuminate/View/Engines/CompilerEngine.php:61",
        "/var/www/html/vendor/facade/ignition/src/Views/Engines/CompilerEngine.php:37",
        "/var/www/html/vendor/laravel/framework/src/Illuminate/View/View.php:139",
        "/var/www/html/vendor/laravel/framework/src/Illuminate/View/View.php:122",
        "/var/www/html/vendor/laravel/framework/src/Illuminate/View/View.php:91",
        "/var/www/html/vendor/osiset/laravel-shopify/src/resources/views/auth/token.blade.php:45",
[...]

PR #956, opened 2021-08-25
Add shop to parameter if we already have it

LonnyX:
This will fix Safari related issue: https://github.com/osiset/laravel-shopify/issues/950

### DIFF
--- a/src/Traits/AuthController.php
+++ b/src/Traits/AuthController.php
@@ -82,7 +82,7 @@ trait AuthController
         if ($query) {
             // remove "token" from the target's query string
             $params = Util::parseQueryString($query);
-            $params['shop'] ??= $shopDomain->toNative() ?? '';
+            $params['shop'] = $params['shop'] ?? $shopDomain->toNative() ?? '';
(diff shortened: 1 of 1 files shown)


osiset:
@LonnyX `??=` is for PHP 7.5 and up only I think? Might be 7.4, but def not 7.3

You'll need to update your code to not use ??= and it'll pass :)

src/Traits/AuthController.php
```diff
@@ -82,8 +82,12 @@ public function token(Request $request)
         if ($query) {
             // remove "token" from the target's query string
             $params = Util::parseQueryString($query);
+            $params['shop'] ??= $shopDomain->toNative() ?? '';
```
osiset, line 85: Use support for 7.3

osiset (changes requested):

src/Traits/AuthController.php
LonnyX: @osiset you're right

osiset (approved):

PR #996, opened 2021-10-14
Update README.md

Dhaval-dinc:


Kyon147:
"plain" is the correct use of language here. So closing this PR as is not needed. Thanks!

Issue #962, opened 2021-09-04
Infinite redirect


sandyrewa:
Install the latest version of the package and also install the latest Laravel but getting continuously redirect and unable to redirect home route after authenticated. Please help how can fix this.

* Package Version: v17.1
* Laravel Version: v8.54
* PHP Version: v8.0

![image](https://user-images.githubusercontent.com/6027540/132094915-82e78e89-88c2-4f65-996b-6734f647fb0b.png)

vlkf:
I got the same problem on fresh install of Laravel and laravel-shopify. It infinite redirects between authenticate/token and home controller.I tried with WAMP and with Laravel built-in server, still the same problem.

- Laravel-Shopify v17.1
- Laravel framework v7.30.4
- PHP v7.3.29

![lara-shopify](https://user-images.githubusercontent.com/81950757/132127379-a931b4fa-d559-465d-beb8-fe692a266ce1.JPG)

vlkf:
Downgrading Laravel-Shopify package to v16.0.6 resolved the problem for me.

sandyrewa:
@vlkf Thanks, it's also working for me with v16.0.6, but I read an article where Shopify reject the app which is created on v16

sandyrewa:
@ClaraLeigh I have done the changes as you mentioned but still getting 302 infinite redirect loops on my local machine but it's working fine server.

ClaraLeigh:
@sandyrewa I don't think its in any of the releases yet, only master. 

To get it working for me, I had to essentially clone the VerifyShopify middleware, and the SessionToken.php, make the changes there and use my custom middleware instead. I think I made a couple of other changes too.... I just had to get it out the door essentially, I've been so busy lately.

I'll probably make some more PRs later but I'm pretty busy till Jan

amosmos:
Thanks @ClaraLeigh! Your solution worked for us :-)
It took us few hours to figure out it is still not in the release, we were sure by now it's there so we didn't check.
I wonder it is not in any release yet.
Amos

sandyrewa:
@amosmos Can you please tell us what exactly needs to be done, it's still not working for us?
Thanks

amosmos:
@sandyrewa 
Do you use one app for many stores with different api_key for each store?
I'm asking because this is the problem that this fix is fixing.
If you only have one api_key then i think you probably have a different problem and that's perhaps why you still have it after making this change.
Amos

sandyrewa:
[...]

PR #1001, opened 2021-11-01
Bugfix incorrect api_key in SessionToken function

ClaraLeigh:
Fixes an issue where the verifyValidity() function checks the api token against the global api_key instead of any shop specific token.
Also fixes an infinite loop error that was occurring for myself, possibly related to: https://github.com/osiset/laravel-shopify/issues/962


osiset:
Thanks I'll have a look.

PR #1005, opened 2021-11-08
:sparkles: Allow currency for billing to be configured

stevesweets:
Means developers in other countries than the US can charge in their native currency.


stevesweets:
Got trigger happy. Incomplete, non-functioning.

PR #1058, opened 2022-01-12
Event to notify Successful App Installation on the Store.

abishekrsrikaanth:
The `AppInstalled` event will be triggered after the installation of App on the store. This can be used to run jobs after successful installation. For example, fetching data from the store that is needed by the application.

Here is an example of how to set up a Listener and subscribe to it on the project's EventServiceProvider.

```php

<?php

namespace App\Listeners;

use Osiset\ShopifyApp\Messaging\Events\AppInstalled;

class AppInstalledEventListener
{
    /**
     * Handle AppInstalled Event.
     *
     * @param  Osiset\ShopifyApp\Messaging\Events\AppInstalled $event
     * @return void
     */
    public function handle(AppInstalled $event)
    {
            $shop = $event->getShop();
            // Get Shop Details
            // Get Products
            // Get Collections
            // etc...
    }
}
```

Once your listener has been defined, you may register it within your application's `EventServiceProvider`:

```php
<?php

namespace App\Providers;

use App\Listeners\AppInstalledEventListener;
use Illuminate\Foundation\Support\Providers\EventServiceProvider as ServiceProvider;
use Osiset\ShopifyApp\Messaging\Events\AppInstalled;

class EventServiceProvider extends ServiceProvider
{
    protected $listen = [
        AppInstalled::class => [
            AppInstalledEventListener::class,
        ],
    ];
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "ext-json": "*",
         "funeralzone/valueobjects": "^0.5",
         "jenssegers/agent": "^2.6",
-        "laravel/framework": "^7.0 || ^8.0",
+        "laravel/framework": "^7.0 || ^8.0 || ^9.0",
         "osiset/basic-shopify-api": "^9.0 || ^10.0"
     },
     "require-dev": {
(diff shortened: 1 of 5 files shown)


talktohenryj:
Cant you just do this with an AfterAuthentication Job? Thats how I do this today.

Kyon147:
@osiset I'm going to close this in favor of keeping the AppInstalledJob and there's a lot of other code changed in here we've already merged in that could make for some messy merging.

Issue #1063, opened 2022-01-21
Custom webhook controller for custom dispatch not used


HugoHeneault:
*For bug reporting only! If you're posting a feature request or discussion, please ignore.*

## Expected Behavior

I'm trying to create a route as stated in [doc](https://github.com/osiset/laravel-shopify/wiki/Creating-Webhooks#optional-custom-controller) but it seems my route it still handled by default webhook controller.

## Current Behavior

I got an error 
```
Too few arguments to function App\Jobs\OrdersPaidJob::__construct(), 2 passed in vendor/laravel/framework/src/Illuminate/Foundation/Bus/Dispatchable.php on line 17 and exactly 3 expected
```

Which seems to demonstrate the dispatch is still made by the default `WebhookController` https://github.com/osiset/laravel-shopify/blob/3ed1a8f3f95cacbdfba65b69857fbc529ecbcb51/src/Traits/WebhookController.php#L29

## Failure Information

Please help provide information about the failure if this is a bug.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Add a route for your webhook in `routes/web.php`: 
```php
Route::post(
    '/webhook/orders-paid',
    [OrderPaidController::class, 'handle']
)->middleware('auth.webhook');

```
2. Create the according controller: 
```php
<?php

namespace App\Http\Controllers;

use App\Jobs\OrdersPaidJob;
use Illuminate\Http\Request;
use Illuminate\Http\Response;
use Osiset\ShopifyApp\Util;
use Illuminate\Http\Response as ResponseResponse;

class OrderPaidController extends Controller
{
    /**
     * Handles an incoming webhook.
     *
     * @param string  $type    The type of webhook
     * @param Request $request The request object.
     *
     * @return ResponseResponse
     */
    public function handle(Request $request): ResponseResponse
    {
        dd('in custom OrderPaidController');
        // Get the job class and dispatch
        $jobData = json_decode($request->getContent());

        OrdersPaidJob::dispatch(
            $request->header('x-shopify-shop-domain'),
            $request->header('x-shopify-webhook-id'),
            $jobData
        )->onQueue(Util::getShopifyConfig('job_queues')['webhooks']);

        return Response::make('', ResponseResponse::HTTP_CREATED);
    }
}

```
3. Call a webhook on the route.
4. You won't get the expected `in custom OrderPaidController` message as result.

### Context

[...]

PR #1066, opened 2022-01-24
Add webhookId to Webhooks

HugoHeneault:
Hi there!

Thanks again for this great laravel & shopify app :)
I missed webhookId value in #1063 so I decided it would be worth adding it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 composer.lock
 .php-cs-fixer.cache
 .phpunit.result.cache
+.idea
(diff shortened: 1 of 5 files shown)


HugoHeneault:
I'm unsure about the `WebhookId|string` types, I just copied how it was already made but don't get why. 🤷‍♂️`

HugoHeneault:
I'm unsure about the `WebhookId|string` types, I just copied how it was already made but don't get why. 🤷‍♂️`

osiset:
@HugoHeneault Actually I am trying to remember if we can use the object values or not due to the serialization of jobs. I think that is why the shop domain is passed in as a string still.

Given that, I think your value object is not needed at this time.

Issue #1071, opened 2022-02-01
### Issue


mudassirali007:
### Issue

If you cancel on the installation page it will not show the install page the second time.

### Reproduce Steps:

1. Goto Oauth Page and select **Cancel**. It will create an entry in the Database in users with the 'remember_token' column as NULL 

<img width="499" alt="Capture" src="https://user-images.githubusercontent.com/54500326/148445662-21db12f3-8406-45be-becc-5ace23fc1547.PNG">

2. Then if you again go to the Oath Page again it will show this message. 

<img width="570" alt="Capture2" src="https://user-images.githubusercontent.com/54500326/148444736-da92d885-3d7d-465f-873d-ad004f34d5d4.PNG">

_Originally posted by @mudassirali007 in https://github.com/osiset/laravel-shopify/issues/614#issuecomment-1006904611_

Kyon147:
I had this issue when setting up a new app for a client. 

From my research (and it was brief) it is actually down to the `deleted_at` column. When you go to the install page and either click cancel or in my case the wrong set-up stack fails before it gets the token and stores it. 

When `Auth` runs again it looks for if a store is deleted it gets a token another token but if `deleted_at` is null/false then it just dies.

I need to set up a PR to test but will let you know.

mudassirali007:
I made a fork with the solution.
check this out [here](https://github.com/osiset/laravel-shopify/pull/1074)

Kyon147:
> I made a fork with the solution. check this out [here](https://github.com/osiset/laravel-shopify/pull/1074)

Thanks for making a PR - I just want to see if we can fix the issue without having to add an additional field etc.

Will do my digging and let you know.

inaryu:
Hi.

First off, let me tell you something. I use a translation app because I'm not very good at English. sorry.

How about the following solution to this issue?

* Modify as follows

```
// src/Http/Middleware/VerifyShopify.php

protected function checkPreviousInstallation(Request $request): bool
{
    $shop = $this->shopQuery->getByDomain(ShopDomain::fromRequest($request), [], true);

    return $shop && ! $shop->trashed(); ← ※ Fixed point
}

↓ Change as follows

protected function checkPreviousInstallation(Request $request): bool
{
    $shop = $this->shopQuery->getByDomain(ShopDomain::fromRequest($request), [], true);

    return $shop && ! $shop->trashed() && !$shop->getAccessToken()->isEmpty(); ← ※ Fix
}
```

[...]

PR #1074, opened 2022-02-04
App Install Page issue

mudassirali007:
fixed this [issue](https://github.com/osiset/laravel-shopify/issues/1071)

### DIFF
--- a/src/Storage/Commands/Shop.php
+++ b/src/Storage/Commands/Shop.php
@@ -74,6 +74,7 @@ class Shop implements ShopCommand
     {
         $shop = $this->getShop($shopId);
         $shop->password = $token->toNative();
+        $shop->installed = 1;
         $shop->password_updated_at = Carbon::now();
 
         return $shop->save();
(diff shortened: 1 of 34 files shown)


src/Storage/Commands/Shop.php
```diff
@@ -74,6 +74,7 @@ public function setAccessToken(ShopIdValue $shopId, AccessTokenValue $token): bo
     {
         $shop = $this->getShop($shopId);
         $shop->password = $token->toNative();
+
```
lucasmichot, line 77: please revert

src/Storage/Commands/Shop.php
```diff
@@ -86,6 +87,7 @@ public function clean(ShopIdValue $shopId): bool
     {
         $shop = $this->getShop($shopId);
         $shop->password = '';
+
```
lucasmichot, line 90: please revert

lucasmichot (changes requested):
See comments
Also can you fix https://github.com/osiset/laravel-shopify/runs/5377785831?check_suite_focus=true ?

Kyon147:
Looks like a fix was already implemented - so closing

Issue #1073, opened 2022-02-02
App must verify the authenticity of the request from Shopify.


somin-parate:
*For bug reporting only! If you're posting a feature request or discussion, please ignore.*

While submitting app we are having those 2 issue 

## Expected Behavior

https://prnt.sc/26ng19g

Please describe the behavior you are expecting.

## Current Behavior

Please describe the current behavior?

## Failure Information

Please help provide information about the failure if this is a bug.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Step 1
2. Step 2
3. ...

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Package Version: 17.1
* Laravel Version: 7.2
* PHP Version: 7
* Using a toolset (Docker, Laradock, Vagrant, etc.):

### Failure Logs

Please include any relevant log snippets or files here.

Kyon147:
I'm not 100% sure that is correct from Shopify.

The first requests look correct for me (this is triggering it from the partner dash but same sort of thing)

![image](https://user-images.githubusercontent.com/32519106/152537225-bf5edde9-2c58-4f36-9beb-e74bed9fb16d.png)

- Click "install" 
- We are sent to the auth route as it has been set up in the app admin.
- We then go to `/admin/oauth/authorize`
- From there it gets redirected to `/admin/oauth/request_grant` where you see the "Install Page"
- Click install and we go to `/admin/oauth/post_grant` to grant the charges (if any)
- Then `/authenticate` happens again with the new `?code` parameter to be authed and all the usual stuff happens i.e we store the API Access Token etc.
- `/authenticate/token` should only happen in between page changes within the app itself not as part of the authentication flow.

Can you provide more detail on your set up with your routes etc.

somin-parate:
@Kyon147  Yes so when we are trying its allowing me to do everything i tried to install that multiple times and its working perfect but still they are sending me same https://wishlist.partners.gemfind.com/login
here is my instance where you can check

Kyon147:
Following the routes via a redirect tracking tool - I don't see /authenticate/token in the flow at all - the only one is when it lands on `/authenticate` and then going to `/oauth/authorize`.

Kyon147:
> @Kyon147 Yes so when we are trying its allowing me to do everything i tried to install that multiple times and its working perfect but still they are sending me same https://wishlist.partners.gemfind.com/login here is my instance where you can check

[...]

PR #1075, opened 2022-02-04
Simplify the auth flow by removing the full page redirect.

Kyon147:
This PR looks to reduce the complexity around the Auth flow but keep the same functionality.

it looks to hopefully solve the issue here around #1073 that because of the "auth/token" coming first before the oAuth it fails the Shopify public app requirements. 

By just redirecting to the pure URL which is built it removes the need to use the `full page` redirect.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
         run: composer install --prefer-dist --no-progress
 
       - name: Normalize composer file
-        if: matrix.normalize == true
+        if: matrix.normalize == true && matrix.php != '8.0'
         run: composer normalize --dry-run
 
       - name: Run test suite
--- a/src/Traits/AuthController.php
+++ b/src/Traits/AuthController.php
@@ -50,7 +50,7 @@ trait AuthController
         if ($status === null) {
             // Show exception, something is wrong
             throw new SignatureVerificationException('Invalid HMAC verification');
-        } else if ($status === false) {
+        } elseif ($status === false) {
             if (! $result['url']) {
                 throw new MissingAuthUrlException('Missing auth url');
             }
--- a/src/Traits/AuthController.php
+++ b/src/Traits/AuthController.php
@@ -50,7 +50,7 @@ trait AuthController
         if ($status === null) {
             // Show exception, something is wrong
             throw new SignatureVerificationException('Invalid HMAC verification');
-        } elseif ($status === false) {
+        } else if ($status === false) {
             if (! $result['url']) {
                 throw new MissingAuthUrlException('Missing auth url');
             }
--- a/src/Traits/AuthController.php
+++ b/src/Traits/AuthController.php
@@ -6,7 +6,6 @@ use Illuminate\Contracts\View\View as ViewView;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Redirect;
-use Illuminate\Support\Facades\Session;
 use Illuminate\Support\Facades\View;
 use Osiset\ShopifyApp\Actions\AuthenticateShop;
 use Osiset\ShopifyApp\Exceptions\MissingAuthUrlException;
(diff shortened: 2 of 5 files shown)


stevesweets:
This makes a dramatic positive difference to the auth flow.

Kyon147:
Hey @osiset - as you are very busy, do you want me to merge this into master and create a new release. This should help a lot of users who are currently getting rejected from Shopify because of the "auth/token" redirect creates in between OAUTH which is against their guidelines. 

Let me know and i can get this out today and set up a patch release.

osiset:
Sure!

stevesweets:
@Kyon147 i think i've discovered an issue here.

If
- you've got an embedded app
- you need to change the scopes

The redirect fails with the old classic `https://yyyyyyy.myshopify.com refused to connect`, and in the console

`Refused to display 'https://yyyyyyy.myshopify.com/' in a frame because it set 'X-Frame-Options' to 'deny'.`

... as we know, and as is referenced here: https://community.shopify.com/c/shopify-apis-and-sdks/changing-scope-x-frame-options/td-p/1193972

Obviously, because if you are in an iframe, one cannot simply:

`
// Just return them straight to the OAUTH flow.
return Redirect::to($result['url']);
`

Any ideas?  Bit baffled by this at the moment, because I think i've also discovered an app bridge bug, and this just happened at the same time.

Kyon147:
Yeah I noticed the same issue, we need to move back to the full page redirect in this instance much like when you change the api scopes the same issue happens.

There's another auth PR which needs to be merged #1097 as this fixes the other issue bust sorts out the iframe redirect one too by moving it back as before.

stevesweets:
This _works_, but is there a particular reason it shouldn't be done this way?

``

return View::make('shopify-app::auth.fullpage_redirect', [
    'authUrl' => $result['url']
]);

``
``
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8">
        <base target="_top">

        <title>Redirecting...</title>

        <script type="text/javascript">
            window.top.location.href = '{{ $authUrl }}'
        </script>
    </head>
    <body>
    </body>
</html>
``

stevesweets:
@Kyon147 apologies, but this seemed to miss your attention.  Do you have any thoughts on if the above would be acceptable?

Kyon147:
Hey @stevesweets 

It's sort of the same result as https://github.com/osiset/laravel-shopify/pull/1097 but this PR uses the AppBridge functions which is probably a little more inline with Shopify if they ever come back on a App Review.

Your way would work in the same essence as the fullpage_redirect view does effectively the same thing.

stevesweets:
> Hey @stevesweets
> 
> It's sort of the same result as #1097 but this PR uses the AppBridge functions which is probably a little more inline with Shopify if they ever come back on a App Review.
> 
> Your way would work in the same essence as the fullpage_redirect view does effectively the same thing.

Ok nice, thanks for the input

talktohenryj:
I've been reading through this. So does this mean if I upgrade to 17.2.0 my app should automatically send merchants to reauthenticate if I change the scopes? 

I ask because I am adding a new scope and when I add it my test store does not automatically reauth. I am currently on 17.1, I am wondering if upgrading to 17.2 will fix my issue. It sounds like it will.

Kyon147:
Hey @talktohenryj - yeah upgrading to 17.2 should fix that issue for you.

talktohenryj:
hmmm, i did the update but still have the issue. Do I need to be on a specific PHP version? I am currently running 7.4 on dev and production.

stevesweets:
> hmmm, i did the update but still have the issue. Do I need to be on a specific PHP version? I am currently running 7.4 on dev and production.

what was your issue, actually?

stevesweets:
> hmmm, i did the update but still have the issue. Do I need to be on a specific PHP version? I am currently running 7.4 on dev and production.

it may be relevant, it may not - but, it will only reauth when the current session ends.

PR #1105, opened 2022-03-10
add variables command

enmaboya:
You will probably add variables to the .env one way or another.

This command just simplifies the this process

### DIFF
--- a/src/Console/AddVariablesCommand.php
+++ b/src/Console/AddVariablesCommand.php
@@ -1,0 +1,117 @@
+<?php
+
+namespace Osiset\ShopifyApp\Console;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Str;
+
+class AddVariablesCommand extends Command
+{
+    /**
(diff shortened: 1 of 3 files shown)


src/Console/AddVariablesCommand.php
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @return array
+     */
+    protected function defaultShopifyVariables(): array
+    {
+        return [
+            'SHOPIFY_APP_NAME' => config('app.name'),
+            'SHOPIFY_API_KEY' => '',
+            'SHOPIFY_API_SECRET' => '',
+            'SHOPIFY_API_SCOPES' => 'read_script_tags,write_script_tags,read_customers,write_customers,read_products,write_products',
```
ItsGageH, line 110: It may be better to have the default scopes as blank. Some users may just leave these 'as-is' which may be flagged by Shopify reviewal as unnecessary scope usage.

Issue #1113, opened 2022-04-01
Shopify AWS EventBridge for Webhook


veed76:
Hi @osiset @Kyon147 @lucasmichot 

I'm registering the webhook via AWS EventBridge, I have added the ARN for the webhook address.
Example:
```
'webhooks' => [
         [
            'topic' => 'app/uninstalled',
            'address' => "arn:aws:events:eu-west-2::event-source/aws.partner/shopify.com/9999999/staging-event-bus"
        ],
],
```

While I'm installing the app in the Shopify store, I got this error: https://prnt.sc/8KxBgePrZSFC

Can you please check and let me where I'm wrong?

Thank you.

Kyon147:
Hi @veed76 

I can't say much for your AWS set up but that does not look like a proper qualified url. 

The error says what the issue is `Invalid URL` so I believe that is your problem.

Kyon147:
> Thanks @Kyon147 for reply. just want to know in this package is it possible to pass AWS ARN in webhook Address?

The webhook needs to be a URL, it's not a requirement of the package but a requirement of Shopify when setting up webhooks using the API.

veed76:
In shopify Documentation they mentioned that we can pass ARN
![image](https://user-images.githubusercontent.com/20658698/161295156-90fa1c7d-8f1c-4c9e-ae6b-9f42abb748ea.png)

veed76:
There is a support for `webhookSubscriptionCreate`. I check in osiste/Laravel Package and there is no support for `eventBridgeWebhookSubscriptionCreate` GraphQL.

lucasmichot:
> There is a support for `webhookSubscriptionCreate`. I check in osiste/Laravel Package and there is no support for `eventBridgeWebhookSubscriptionCreate` GraphQL.

Hi @veed76 , indeed there is no support.
Feel free to make a PR or just ask a Shopify developer to do it

Kyon147:
Interesting, good testing -it looks like `webhookSubscriptionCreate` uses type `URL` https://shopify.dev/api/admin-graphql/2022-01/scalars/URL

Which is why `eventBridgeWebhookSubscriptionCreate` exists to be able to use the ARN routes. 

As @lucasmichot - the package does not have that GraphQL support in it, as you've found yourself. So feel free to provide a PR.

veed76:
Hi @lucasmichot @Kyon147 
[...]

PR #1114, opened 2022-04-02
Added the support for AWS EventBridge Shopify Webhook

veed76:
There is support for webhookSubscriptionCreate. I check in osiste/Laravel Package and there is no support for eventBridgeWebhookSubscriptionCreate GraphQL.

So I have added the support for eventBridgeWebhookSubscriptionCreate

Issue: https://github.com/osiset/laravel-shopify/issues/1113

### DIFF
--- a/src/resources/config/shopify-app.php
+++ b/src/resources/config/shopify-app.php
@@ -328,6 +328,22 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Shopify Webhooks Address Type
+    |--------------------------------------------------------------------------
+    |
+    | This option is for defining webhooks type.
+    | `callbackUrl` it's use https:// route.
+    | `arn` it's use arn route.
+    |
+    | This documentation will helps you
+    | https://shopify.dev/apps/webhooks/configuration/eventbridge
+    |
+    */
+    
+    'webhook_address_type' => env('SHOPIFY_WEBHOOK_ADDRESS_TYPE', 'arn'),
+
+    /*
+    |--------------------------------------------------------------------------
     | Shopify Webhooks
     |--------------------------------------------------------------------------
     |
--- a/src/Services/ApiHelper.php
+++ b/src/Services/ApiHelper.php
@@ -366,26 +366,44 @@ class ApiHelper implements IApiHelper
      */
     public function createWebhook(array $payload): ResponseAccess
     {
-        $query = '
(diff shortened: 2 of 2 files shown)


src/resources/config/shopify-app.php
```diff
@@ -326,6 +326,22 @@
...
+    | This option is for defining webhooks type.
+    | `callbackUrl` it's use https:// route.
+    | `arn` it's use arn route.
+    |
+    | This documentation will helps you
+    | https://shopify.dev/apps/webhooks/configuration/eventbridge
+    |
+    */
+    
+    'webhook_address_type' => env('SHOPIFY_WEBHOOK_ADDRESS_TYPE', 'arn'),
```
Kyon147, line 343: Can you please make `callbackUrl` the default. As a lot of current users will be using the standard webhook approach this allows for the easiest backwards compatibility.

Kyon147 (changes requested):
Great PR but some minor feedback.

src/resources/config/shopify-app.php
veed76: Hi @Kyon147 I have updated to `callbackUrl ` please check.

Thank you.

veed76:
Hi @Kyon147 
Can you please review the changes and merge it ASAP, so I can use this in the new version.

Thanks.

Kyon147:
@veed76 - can you take a look at your code coverage as it has dropped a lot 80% down from 92.70%

veed76:
@Kyon147 Can you please tell me next what I need to do?

shota:
@veed76 
Amazing PR. This is really what we want! However, checking address type with `webhook_address_type` configuration variable sometimes have another problem when some user want to mix EventBridge and REST webhook handling. 

How about checking webhook URL like below?

[src/Services/ApiHelper.php](https://github.com/osiset/laravel-shopify/pull/1114/files#diff-c7040d0b15fabd2e480e0b6864158fd6d08f374e8ff2ebc67c009912bbb828cf) line 370
```
if ($addressType === 'arn') {
```

To new one like ... 
```
if (str_starts_with($payload, 'arn:')) {
```

This implementation is much more like old version which support EventBridge URL.

osiset:
Hi @veed76
Essentially you need to edit the existing tests for ApiHelper, to ensure if someone changed the config, then the proper code is called.

PR #1100, opened 2022-03-03
Adding support for Laravel 9

steveperrycreative:

PR #1120, opened 2022-04-15
Added laravel 9

isyedumer:
Updated composer.json to support laravel 9


Kyon147:
Thanks for opening a PR but there's one already for this. https://github.com/osiset/laravel-shopify/pull/1100

PR #967, opened 2021-09-05
Fix getDomain() error and throw catchable exception

therealdb:
Resolves #912

PR #1121, opened 2022-04-16
Add login page for manual login of a store

adrianmihaila:


Kyon147:
Just to update a new release will be coming later on v17.1.1 which will have this PR 

https://github.com/osiset/laravel-shopify/pull/967

We will throw a proper Missing  Domain Exception that you can handle how you like.

PR #1097, opened 2022-03-01
Fixed app-bridge redirect on fullpage_redirect auth view

olavoasantos:
## Summary

Shopify is deprecating EASDK and the package was making making an EASDK calls (ie `Shopify.API.remoteRedirect`). This triggered the message for the apps consuming the library.

> Closes #1094

## What was done?

This PR refactors the EASDK call to redirect the app made on the `auth/fullpage_redirect.blade.php` view to use App Bridge instead.

PR #1139, opened 2022-05-19
Fix when we try to reauth scopes

Kyon147:
This PR brings back the full page reauth because when you update the api scopes after install you will run into the iframe error as it is set to deny.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 composer.lock
 .php-cs-fixer.cache
 .phpunit.result.cache
+/.idea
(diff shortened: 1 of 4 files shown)


Kyon147:
@osiset I found an issue with my previous implementation that happens when you try to reauth an app. Usually when requesting new permissions after the first install.

I've brought back the full page redirect but updated it to use the app bridge implementation.

Kyon147:
As a side note but connected to this. 

We do need to fix the OAUTH flow as the current way where you hit /auth/token after accepting the install gets flagged in review for public apps as they expect the merchant to hit the home route (UI) right after and are very strict with it.

image

Kyon147:
Closing in favor of https://github.com/osiset/laravel-shopify/pull/1097

PR #1142, opened 2022-05-22
Verify theme support

enmaboya:
As you know lately, Shopify does not allow applications that only use script tags. 

To work with the Online Store 2.0 you need to use "Theme app extensions".
But in this case you should not install script tags.
If the store has not yet updated the theme and is using the Online Store 1.0, script tags must be installed.

These changes solve this problem


Kyon147:
@enmaboya can you take a look at the tests for this PR too. There's a drop in coverage of about 4%.

abishekrsrikaanth:
@enmaboya Great work on this PR. Any chance you can brief on what the VerifyThemeSupport class does?

PR #1145, opened 2022-05-24
Propose to fix Billable app throwing exception on install

thang12l:
Fix PHP 8.0: Deprecate required parameters after optional parameters in function/method signatures


Kyon147:
@osiset going to merge this one (later) in too for the upcoming release as php8.1 is being used more and more and this fix stops the billing route breaking as the optional param needs to be at the end of the method.

Kyon147 (approved):

PR #1172, opened 2022-07-24
Add variables command

enmaboya:
You will probably add variables to the .env one way or another.

This command just simplifies the this process

### DIFF
--- a/src/Console/AddVariablesCommand.php
+++ b/src/Console/AddVariablesCommand.php
@@ -84,7 +84,7 @@ class AddVariablesCommand extends Command
         return $this->option('force')
             ? true
             : $this->confirm(
-                "This will invalidate $key variable. Are you sure you want to override $key?"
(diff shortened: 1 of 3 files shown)


osiset:
@enmaboya I am fine with this! Maybe the command should be `shopify-app:variables:generate`? Been looking at how other pkgs do commands, seem to be more along the lines of `namespace:section:action`.

Approving running the actions too, so we can see test results.

osiset (approved):

osiset:
Hmm not sure why actions is not showing up for this PR... @Kyon147 have you seen this happen before?

Kyon147:
@osiset that is strange, we might need to do a blank commit and push in this PR to trigger them. Github might have just had a wobble.

osiset:
@Kyon147 That solved it :/

osiset:
@enmaboya Merged. Maybe add a wiki entry about it if you get time to explain what it does... we probably need a page to explain all commands existing (like this one and webhook) in the future.

PR #1178, opened 2022-07-28
Setting up Iframe protection

enmaboya:
At the moment, if you do not set some of the headers, the application will not pass the review.

This header (frame-ancestors) allows you to open the application in an iframe only from an allowed domain

### DIFF
--- a/src/Http/Middleware/IframeProtection.php
+++ b/src/Http/Middleware/IframeProtection.php
@@ -5,8 +5,8 @@ namespace Osiset\ShopifyApp\Http\Middleware;
 use Closure;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Cache;
-use Osiset\ShopifyApp\Objects\Values\ShopDomain;
 use Osiset\ShopifyApp\Contracts\Queries\Shop as IShopQuery;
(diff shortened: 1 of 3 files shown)


Kyon147:
@enmaboya if you pull in master, it should fix the issue with the allowed plugins error but you'll need to fix the linting issue still.

enmaboya:
@Kyon147 linting has been updated

Kyon147:
Thanks @enmaboya 

Could you also write a test for your new code, as the coverage drops by 1% due to the file being included.

image

We should be good after a test it written that covers your new code.

enmaboya:
@Kyon147 test added

Kyon147:
Thanks for your work on this @enmaboya 👍

@osiset can you take a look at approve too, it will be good to get this out with laravel 9 as this blocks app store approvals at the moment without the iframe protection.

osiset:
@Kyon147 Just reviewed it, I am good with it to merged in!

Kyon147:
Thanks @osiset will sort it out later to master and it can be part of the next release candidate for laravel 9 support. Will sort v17.2.0 next week so we can have the new version out.

Issue #1174, opened 2022-07-25
When a user declines a charge, the return URL has no chargeId, causing 500 error


usmanpakistan:
## Expected Behavior

When a user decline a charge, he is redirected to Shopify Settings page as charge decline is handled by Shopify since 2018. When a user closes the setting popup, it should be redirected to App home page

## Current Behavior

Problem is, on closing the popup, the user is redirected to the return URL, i.e. billing/process/{plan_id}/?shop=myshop.myshopify.com. The return URL, in case of decliced charge, does not contain the charge ID. This causes the route to show an error. 

## Failure Information

Billing process page shows an error if chargeId is missing from the query string

### Steps to Reproduce

1. Install a new app or open the change plan screen of  your app
2. Select a different plan. You will be redirected to Shopify Charge page
3. Do not approve the Charge and decline the charge by clicking on the cancel button.
4. You will be redirected to Shopify admin, and will see the new setting popup.
5. Close the popup, you will be redirected to /billing/process/{plan_id} page of the app
6. The page shows 500 error

### Context

* Package Version: v17.1.0
* Laravel Version: v8.76
* PHP Version: v7.4

### Failure Logs

See the screencast here
https://gyazo.com/fae78b8d3653141dbdd0fcabef63924c

usmanpakistan:
@osiset @Kyon147 Can you people look into this. Thanks

Kyon147:
Closing this now it has been merged, thanks for the PR!

PR #1181, opened 2022-08-02
Update BillingController.php

usmanpakistan:
Added a check in billing controller to redirect to the home route if no charge ID is provided. Fix issue #1174

### DIFF
--- a/src/Traits/BillingController.php
+++ b/src/Traits/BillingController.php
@@ -74,7 +74,10 @@ trait BillingController
     ): RedirectResponse {
         // Get the shop
         $shop = $shopQuery->getByDomain(ShopDomain::fromNative($request->query('shop')));
-
+        if(!$request->has('charge_id'))
+            return Redirect::route(Util::getShopifyConfig('route_names.home'), [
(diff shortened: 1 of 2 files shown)


Kyon147:
Thanks for looking into this @usmanpakistan it was on my list this week so was happy to see someone else providing a solution 💪 

i've run the tests and a couple are failing, can you please take a look and fix. One of them is a composer allowed plugins, which probably just needs to be added into the package.json as an allowed plugin.

Kyon147:
My PR #1194 should fix the issue with the normalize plugin.

Kyon147:
if you pull in master, it should fix the issue with the allowed plugins error but you'll need to fix the linting issue still.

Kyon147:
@osiset - this is the other breaking change which will get an app rejected.  @usmanpakistan got part the way there but have finished it off and added a test so i can merge this into master today (if you can take a look) and then it can go with release v17.2.0

osiset:
@Kyon147 Will look in an hour, thanks

osiset (approved):
@Kyon147 All good.

PR #1192, opened 2022-08-25
Allow for custom "shops" table for automatic migrations

nahid:
laravel-shopify allowing custom auth guard in their latest version. But it not support to allowing migration with custom table for shops in automatic migration. 

This PR is allowing developer's to set their own shops table in config to execute migration with given name.

### DIFF
--- a/src/Storage/Commands/Charge.php
+++ b/src/Storage/Commands/Charge.php
@@ -3,6 +3,7 @@
 namespace Osiset\ShopifyApp\Storage\Commands;
 
 use Illuminate\Support\Carbon;
+use Illuminate\Support\Str;
 use Osiset\ShopifyApp\Contracts\Commands\Charge as ChargeCommand;
 use Osiset\ShopifyApp\Contracts\Queries\Charge as ChargeQuery;
 use Osiset\ShopifyApp\Objects\Enums\ChargeStatus;
@@ -50,10 +51,12 @@ class Charge implements ChargeCommand
             return $obj instanceof Carbon;
         };
 
+        $userTableId = Str::singular(Util::getShopifyConfig('table_names.shops', 'users')) . '_id';
+
         $chargeClass = Util::getShopifyConfig('models.charge', ChargeModel::class);
         $charge = new $chargeClass();
         $charge->plan_id = $chargeObj->planId->toNative();
-        $charge->user_id = $chargeObj->shopId->toNative();
+        $charge->$userTableId = $chargeObj->shopId->toNative();
         $charge->charge_id = $chargeObj->chargeReference->toNative();
         $charge->type = $chargeObj->chargeType->toNative();
         $charge->status = $chargeObj->chargeStatus->toNative();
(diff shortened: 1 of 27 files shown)


src/Storage/Commands/Charge.php
```diff
@@ -89,10 +92,11 @@ public function delete(ChargeReference $chargeRef, ShopId $shopId): bool
      */
     public function makeUsage(UsageChargeTransfer $chargeObj): ChargeId
     {
+        $userTableId = Str::singular(Util::getShopifyConfig('table_names.shops', 'users')) . '_id';
```
osiset, line 95: @nahid  This line is used in multiple spots in multiple files, I think it might be best to make a new method in Util to get this value.

osiset: Maybe something like this:

```php
# Util file
public static getShopsTable(): string {
  return Util::getShopifyConfig('table_names.shops', 'users'));
}
```

Usage like:

```php
Util::getShopsTable()

# or

$shopsTable = Util::getShopsTable();
$userId = "{$shopsTable}_id";
$planForeign = "{$shopsTable}_plan_id";
```

osiset (changes requested):
Good update, comments inside about a way to clean things up.
Can re-review after.

src/Storage/Commands/Charge.php
nahid: Thanks, @osiset for your reply. Yes, it would be better to reuse if we separate this code. I'll do this ASAP.

nahid:
@osiset, I have already rewritten the code based on your change request. Please review the PR again. Thank you

osiset:
@nahid I agree. Go ahead with it :)

nahid:
@osiset, I have pushed the following changes, Please review again

osiset:
Solid. I'll approve letting Actions run and we'll see what happens. Probably will need test added to Until tests to ensure coverage is there.

nahid:
> Solid. I'll approve letting Actions run and we'll see what happens. Probably will need test added to Until tests to ensure coverage is there.

Maybe `tests/Storage/Queries/ChargeTest.php` file needs to be updated to code coverage. Please let me know if there is anything I can do. Thank you

osiset:
@nahid Tests pass, coverage didnt change as well, so thats OK.
Linting failed though, please see the Action for details.

nahid:
> @nahid Tests pass, coverage didnt change as well, so thats OK.
> Linting failed though, please see the Action for details.

@osiset, I just pushed another commit with fixes, please check

osiset (approved):

osiset:
@nahid Still is failing the lint.

Locally, you can `vendor/bin/php-cs-fixer fix --diff --dry-run` to see what needs to be changed, and `vendor/bin/php-cs-fixer fix` if you're OK with automatic changes.

nahid:
@osiset please check the latest commits

osiset (approved):

Kyon147:
> Tests now pass, linting is good. @nahid 
> 
> @Kyon147 I'm good with this one too if you are, seems like a feature thats been asked about before a few times.

Yeah looks good to me too

osiset:
@nahid Please review once more, and if you're good we'll merge this in.

nahid:
> @nahid Please review once more, and if you're good we'll merge this in.

@osiset sure

PR #1194, opened 2022-08-26
Allow ergebnis/composer-normalize to run the code clean on workflows

Kyon147:
`ergebnis/composer-normalize` is being stopped by composer to normalize because it is not an allowed plugin and this is killing the tests in the workflows.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,10 @@
         "phpunit/phpunit": "~8.0 || ^9.0"
     },
     "config": {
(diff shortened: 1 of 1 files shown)


Kyon147:
@osiset - this is a fix to get tests up and running again now composer is breaking the tests because the normalizer is not in the allowed plugins.

osiset:
Good with this, merged. Thanks!

PR #1207, opened 2022-09-09
Deprecation supports and fixes

osiset:
* Fixes deprecation notices for PHP8
* Adds deprecation logging support to tests being run

Mainly, `parse_url` and `parse_str` calls have been updated. Warnings were present complaining about possible null values where a string is expected.

This PR adjusts to check for falsey/null and use a empty string as fallback.


Kyon147 (approved):
Looks good 👍

osiset:
Merged. Closed.

PR #1220, opened 2022-09-26
Adding event to notify App Installation on the store

abishekrsrikaanth:
The `AppInstalled` event will be triggered after the installation of App on the store. This can be used to run jobs after successful installation. For example, fetching data from the store that is needed by the application.

Here is an example of how to set up a Listener and subscribe to it on the project's EventServiceProvider.

```php

<?php

namespace App\Listeners;

use Osiset\ShopifyApp\Messaging\Events\AppInstalled;

class AppInstalledEventListener
{
    /**
     * Handle AppInstalled Event.
     *
     * @param  Osiset\ShopifyApp\Messaging\Events\AppInstalled $event
     * @return void
     */
    public function handle(AppInstalled $event)
    {
            $shop = $event->getShop();
            // Get Shop Details
            // Get Products
            // Get Collections
            // etc...
    }
}
```

Once your listener has been defined, you may register it within your application's `EventServiceProvider`:

```php
<?php

namespace App\Providers;

use App\Listeners\AppInstalledEventListener;
use Illuminate\Foundation\Support\Providers\EventServiceProvider as ServiceProvider;
use Osiset\ShopifyApp\Messaging\Events\AppInstalled;

class EventServiceProvider extends ServiceProvider
{
    protected $listen = [
        AppInstalled::class => [
            AppInstalledEventListener::class,
        ],
    ];
}
```

@Kyon147, I am resending this PR again as the last PR (#1058) had additional code that was already merged in to the repo.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "ext-json": "*",
         "funeralzone/valueobjects": "^0.5",
         "jenssegers/agent": "^2.6",
-        "laravel/framework": "^7.0 || ^8.0 || ^9.0",
+        "laravel/framework": "^7.0 || ^8.0 || ^9.0 || ^10.0",
         "osiset/basic-shopify-api": "^9.0 || <=10.0.5"
     },
     "require-dev": {
--- a/src/Actions/InstallShop.php
+++ b/src/Actions/InstallShop.php
@@ -5,6 +5,7 @@ namespace Osiset\ShopifyApp\Actions;
 use Exception;
 use Osiset\ShopifyApp\Contracts\Commands\Shop as IShopCommand;
 use Osiset\ShopifyApp\Contracts\Queries\Shop as IShopQuery;
+use Osiset\ShopifyApp\Messaging\Events\AppInstalled;
 use Osiset\ShopifyApp\Objects\Enums\AuthMode;
 use Osiset\ShopifyApp\Objects\Values\AccessToken;
 use Osiset\ShopifyApp\Objects\Values\NullAccessToken;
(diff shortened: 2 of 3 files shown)


Kyon147:
Hi @abishekrsrikaanth 

Thanks for submitting a PR. Could you let me know how you would use this differently to the `AfterAuthenticate` job that already exists in the package which can be used for the same thing.

Thanks!

nahid:
Also, need to add the `AppUninstalled`, `PlanPurchased`, and `PlanChanged` events.

@osiset and @Kyon147

nahid:
@osiset Thanks, got it. I think the `AfterAuthenticate` job feature should be deprecated in the next version but legacy support is required up to a certain version. It can be configurable. If someone wants to use the legacy feature he/she can't configure anything but if someone is want to use the Event Listener feature they need to configure it.

> In my previous comment I meant that we need to add `AppUninstalled`, `PlanPurchased`, and `PlanChanged` events besides the `AppInstalled` event. Because sometimes App owners need to notify the shop owners of these events.

nahid:
Thank you @osiset, I'll make a PR with these changes ASAP.

osiset:
@nahid One thing to note for appinstalled, is that there will need to be an option to run it "inline" and not later somehow. Some people require changes to be made before the shop user sees the app (example, set some flag in the database).

bilfeldt:
> @nahid One thing to note for appinstalled, is that there will need to be an option to run it "inline" and not later somehow. Some people require changes to be made before the shop user sees the app (example, set some flag in the database).

@osiset Given that this is not a job being dispatch but an event, then note that events are always synchronous. The listeners one might register for these events on the other hand can be either queued or synchronous. So "inline" is essentially build into the implementation using events, making it a really nice solution for any use case 👍

PR #1222, opened 2022-09-28
Removed `!!` from the full_page_redirect breaking authenticate route

Kyon147:
Currently the authenticated route can break usually on a proper visit to /authenticate because of the `{{!! !!}}` formatting in the blade template for the api key, shop origin and host.

### DIFF
--- a/src/resources/views/auth/fullpage_redirect.blade.php
+++ b/src/resources/views/auth/fullpage_redirect.blade.php
@@ -23,9 +23,9 @@
                     var createApp = AppBridge.default;
                     var Redirect = AppBridge.actions.Redirect;
                     var app = createApp({
-                        apiKey: "{{!! $apiKey !!}}",
-                        shopOrigin: "{{!! $shopOrigin !!}}",
-                        host: "{{!! $host !!}}",
+                        apiKey: "{{ $apiKey }}",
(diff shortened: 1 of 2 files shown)


Kyon147:
@osiset a little bit of a hotfix as the auth route is currently broken when accessing it directly.

osiset:
@Kyon147 It seems [testbench-core](https://github.com/orchestral/testbench-core) introduces static properties. Specifically [here](https://github.com/orchestral/testbench-core/blob/7.x/src/TestCase.php#L50), but its not declared in the class head, resulting in and error for PHP 8.

osiset (approved):

Kyon147:
> It was literally added 10 hours ago. Seems like its trying to set lastResponse, but its not defined anywhere. Do we need to define it in TestCase.php of our package? Like... `public static $latestResponse = null;` ?

Yeah that is totally weird, I'll take a look but defining it in our test case sounds like a good idea for now!

Kyon147:
@osiset that has fixed the error for now in the tests - not sure why codecov is complaining about files not even touched in my PR though....

osiset (approved):
@Kyon147 Good to go, this will fix up other PR tests for PHP8. Codecov is fine :)

PR #1235, opened 2022-10-12
Added parameter to billing process

RishabhTayal:
When a merchant accepts/declines billing this will add a parameter 'billing -> success' to the home page route. This way developers can detect when billing is accepted/declined.

### DIFF
--- a/src/Traits/BillingController.php
+++ b/src/Traits/BillingController.php
@@ -92,6 +92,7 @@ trait BillingController
         // Go to homepage of app
         return Redirect::route(Util::getShopifyConfig('route_names.home'), [
             'shop' => $shop->getDomain()->toNative(),
+            'billing' => $result ? 'success' : 'failure'
         ])->with(
             $result ? 'success' : 'failure',
             'billing'


osiset:
@RishabhTayal Good idea, thanks for this. Can you add a test to the BillingController's test file to confirm this works as expected? You can modify the existing tests thats in the file already for accept/decline and test the result of the redirect URL.

osiset:
@RishabhTayal You should be able to assert the parameter comes back on [this test](https://github.com/RishabhTayal/laravel-shopify/blob/billing-parameter/tests/Traits/BillingControllerTest.php#L53)

RishabhTayal:
@osiset Sorry for naive question: Do you have an example in the code of asserting a parameter?

Kyon147:
@osiset I'm assuming this PR is the one you mentioned, I can merge this one and my on together once they are both ready and they can go out together.

osiset:
@Kyon147 This is exactly it, yes!

Kyon147:
Closing in favour of #1244

PR #1244, opened 2022-10-26
AppBridge : Add the host param to the billing redirect url

Kyon147:
Currently when trying to use AppBridge within an SPA, if we leave the SPA context like when billing happens we need to not only pass the shop param but also the new host param that shopify uses to load AppBridge.


Kyon147:
Hey @osiset this is ready for your review.

osiset (approved):
@Kyon147 Good to go!

PR #1247, opened 2022-10-27
Updates to token_handler.blade.php to support Livewire 2.x

### LINKED ISSUE #1245: Alpine.js/Livewire and Power-Grid Authentication redirect hangup
*For bug reporting only! If you're posting a feature request or discussion, please ignore.*

## Expected Behavior

When making a a request via Livewire, after adding the auth header, we expect the requested resource to load.

## Current Behavior

When queries are fired, the appropriate header is added. However, the middleware "verify.shopify" is forcing a redirect through responding with HTML and trying to redirect from javascript. This cause the request to end and errors abound. 

This does not work...

### Steps to Reproduce

1. Use livewire

unplugged216:
Currently there is no support for Livewire 2.x. This PR provides support for Livewire 2.x by incorporating the required headers into `token_handler.blade.php`.

This closes #1245

### DIFF
--- a/src/resources/views/partials/token_handler.blade.php
+++ b/src/resources/views/partials/token_handler.blade.php
@@ -31,6 +31,15 @@
                 window.jQuery.ajaxSettings.headers = { 'Authorization': bearer };
             }
         }
+		
+		if (window.Livewire) {
+            // livewire
+            window.livewire.addHeaders({
+                'Authorization': bearer,
+                'content-type': 'application/json',
+                'X-Requested-With': 'XMLHttpRequest'
+            });
+        }
 
         if (window.axios) {
             // Axios


bilfeldt:
@unplugged216 awesome 👍 

## Authentication using token

How does Livewire authenticate these requests? [By default Livewire uses the `web` middleware group](https://github.com/livewire/livewire/blob/65d311c63d2d1092ad5891fd0c1c1bdbed272cee/config/livewire.php#L84) but it also [adds some extra middlewares to each request](https://github.com/livewire/livewire/blob/65d311c63d2d1092ad5891fd0c1c1bdbed272cee/src/LivewireManager.php#L24) including the `\Illuminate\Auth\Middleware\Authenticate::class` one. But this middleware would per default use session based authentication and we do not have that 🤔 

So am I missing something or would we need to somehow tell Livewire to authenticate using the session token provided as a bearer token?

## Refresh of token

The session token is per default only valid for 60 seconds and it's recommended to refresh it every 2sec. Would we not need to implement a function that returns the session token if it's still valid or creates a new one if it has expired.

Otherwise any livewire request made 60 sec after the page has loaded would fail? 🤔

unplugged216:
Great questions @bilfeldt !

**Authentication**
Livewire relies on whatever middleware you place inline. This can be configured in `config/livewire.php` for the `middleware_group` variable. You can modify it to include `verify.shopify` similar to this: `'middleware_group' => ['web','verify.shopify']`. This is how I am currently doing it. 

**Refresh**
Nothing changed here. As depicted in the code for the commit, this addition is inline with the other supported frameworks, vuejs and jQuery. 
[Commit](https://github.com/osiset/laravel-shopify/commit/31ff8ea3698ac849d22d902f045ea938ef211fc5)

Hopefully that answered your questions.

bilfeldt:
Thanks so much for taking the time to explain @unplugged216 👍 

## Authentication

It makes total sense that Livewire is capable of authenticating the request if you add the `verify.shopify` middleware. I was not aware that the config `middleware_group ` can also accept single middlewares and not just a group (somewhat misleading naming there) but this is clearly what you should do according to the docs: https://laravel-livewire.com/docs/2.x/authorization#introduction

## Refresh

The refreshing is implemented using the [`keepRetrievingToken` method](https://github.com/osiset/laravel-shopify/blob/695f968859be1bf041c5b1df8d67640502e5ce1a/src/resources/views/partials/token_handler.blade.php#L8) I can see.

Issue #1250, opened 2022-10-31
"Shopify App Outdated" but running latest app bridge, osiset, etc


talktohenryj:
*For bug reporting only! If you're posting a feature request or discussion, please ignore.*

## Expected Behavior

When going to the app it should go directly to the home page.

## Current Behavior

When going into the app, I receive this message. 

![Frame 58](https://user-images.githubusercontent.com/88204834/199120422-6eae3509-8fa6-41f0-aa6f-a08bb96a896f.png)

## Failure Information

This happens every time I try to log into my app on production. It does not show on development. When I click the link provided, it says I am outdated because I am not compatible with the new Shopify admin domain. I don't believe that is true. Below is some additional info.

### Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Package Version: v17.2
* Laravel Version: v8.83.25
* PHP Version: v7.4
* 
### Failure Logs

I don't have much for failure logs but I am running the latest app bridge as set by Osiset.
My app is a multipage application. That info may help.

Also, after speaking with a friend he believes the issue may be: "If your app isn't a [single-page application](https://developer.mozilla.org/en-US/docs/Glossary/SPA), then your server needs to persist the host value during the initial page load. On subsequent page loads, you should use the persisted host value to configure App Bridge."

If the above is the case is there a way in Osiset to set the host value as persistent? We tried to use cookies but were unsuccessful. Also, not 100% sure that is the problem.

Any advice would be helpful. This is a terrible notification from Shopify and I believe really hurting my installation rate right now.

jhaineymilevis:
same problem

talktohenryj:
I messaged Shopify as well so hopefully we'll get an answer soon.

nikwen:
https://github.com/osiset/laravel-shopify/issues/1248 might be related.

jhaineymilevis:
@talktohenryj  have you fixed it?

CedricVleminckx:
Same problem in a Single page application.

Shopify says our host parameter is configured wrong in app bridge.

We also opened multiple support tickets to request a dev store on the new admin.shopify.com domain, but they say they can not help us. So there is no way for us to test this :(

This was the latest response from Shopify:
when we tried to open that last request to your /authenticate/token path in a new browser tab, it takes us to the legacy/store domain and They are getting a 500 error from your app if they don’t remove that host value attached to the target parameter, which is for redirection after authenticating.

[...]

PR #1260, opened 2022-11-09
Issue 1250 fix

jhaineymilevis:

PR #1261, opened 2022-11-10
fix issue #1250

bugfixman:
Now the host parameter is saved in the session and available anywhere


Kyon147:
Please don't duplicate PR's and this has not updated tests. Closing in favour of the currently open #1260

